Issue #65, opened 2021-03-01
[Feature] Support allowed argument values as an option


abh1kg:
First of all, I cannot sing the praises of `bashly` enough - I have been using it heavily for a BASH CLI in my project for quite some time now and it _just works_.

One feature that I'd like to see personally is support for allowed options for arguments. For example, a suggestion for the `bashly.yml` could look like:

```
- name: login
  short: l
  help: Log on to a cool service

  args:
  - name: role
    help: Role for user
    required: true
	allowed: ["admin", "user"]
```

The behaviour I am expecting is:
:heavy_check_mark: `login admin`
:heavy_check_mark: `login user`
:x: `login anythingelse` - this could fail with an error message like `Invalid role - allowed values are "admin" and "user"`

DannyBen:
I am happy to hear that other people make good use of bashly, thanks for sharing that.

While I can see the value of what you are proposing, my initial reaction is that this is something that is better handled by the application space, rather than the framework space.

### Why I think it should be handled by the user-domain application

Note that even in your question, you wanted to show a friendly error to the user:

> `Invalid role - allowed values are "admin" and "user"`

So, if this kind of feature is handled by the framework, it will require doing quite a few things. Such as, handle the help message accordingly, perhaps alter the usage text, add the same kind of feature to the positional arguments, and possibly more.

Also keep in mind, one of bashly's objectives is to be easy to learn, and easy to configure. I intentionally created a minimal set of tools - so much so, that some "features" were excvluded and can be opt in by calling `bashly add FEATURE`.

### How I propose to do it (without any change to bashly).

I think you can still implement what you want, with ease, and without "copy pasting" code around. You can even do so in a way that is easy to then copy the solution between your different scripts.

Here is what I suggest:

#### Step 1: Initialize a script workspace with functions lib

```shell
$ bashly init --minimal
$ bashly add lib   # this just creates a sample function in `./sec/lib`
```

#### Step 2: Update bashly.yml with a sample flag+argument

```yaml
# bashly.yml
name: login
help: Sample minimal application without commands
version: 0.1.0

args:
- name: site
  help: Site to login to

flags:
- long: --user
  short: -u
  arg: NAME
  help: "User name. Allowed: user, admin"
```  

#### Step 3: Create the `verify_user` function

```bash
# src/lib/verify_user.sh
verify_user() {
  allowed="$1"
  if [[ ! ${args[--user]} =~ $allowed ]]; then
[...]

PR #68, opened 2021-03-02
Add support for whitelist in args

### LINKED ISSUE #67: Support whitelist for args
Similarly to what is done in #66 and discussed in #65 , we should probably implement the same for args.

DannyBen:
closes #67, cc #65

---

TODO:

- [x] Implementaiton
- [ ] Second opinion
- [x] Example
- [x] Specs
- [x] README

---

This PR adds a whitelist for arguments, and unifies the arg and flag whitelist implementation.


DannyBen:
@abh1kg - this branch contains the updated flag whitelist implementation, and the new arg whitelist.
Can you check that it works as advertised?

There is an example [here](https://github.com/DannyBen/bashly/tree/add/arg-whitelist/examples/whitelist) - see the bashly.yml in it. In particular, we need to pay attention to how they both behave when there is a default value or `required: true`. I think I covered it all from testing standpoint (the test.sh in that example folder shows the commands I am running).

DannyBen:
@abh1kg should I wait for you to test, or release it?

Issue #74, opened 2021-06-01
catch_all arguments are not printed if not any arg is specified


RonnyLV:
`catch_all` arguments are not printed if not any arg is specified.

Expected generated help for `catch_all` `label` and `help` even when `args` is not filled in.

DannyBen:
Can you provide the smallest YAML that reproduces the problem?

RonnyLV:
```yaml
name: cli
help: Sample application
version: 0.1.0

commands:
- name: download
  short: d
  help: Download a file
  catch_all:
    label: "URLs"
    help: "List of URLs"

PR #75, opened 2021-06-01
Fix "catch_all arguments are not printed if not any arg is specified #74"

RonnyLV:


DannyBen:
Thanks for the report and PR.
I added a spec for it, and will release a gem version shortly.

PR #109, opened 2021-09-30
Add `bashly generate --quiet`

wolfgang42:
For development, I have a script that wraps my command and runs `bashly generate` before each invocation to pick up any changes I've made. In this situation, the output of the generation is extraneous and distracting, so I've added a `--quiet` switch which hides all the boring changes (though it does still print `created #{file}` after generating any new files).

### DIFF
--- a/spec/bashly/commands/generate_spec.rb
+++ b/spec/bashly/commands/generate_spec.rb
@@ -51,6 +51,34 @@ describe Commands::Generate do
     end
   end
 
+  context "with --quiet" do
+    let(:cli_script) { "#{target_dir}/cli" }
+
+    before do
+      reset_tmp_dir
+      success = system "mkdir -p #{source_dir} && cp lib/bashly/templates/bashly.yml #{source_dir}/bashly.yml"
+      expect(success).to be true
+    end
+
+    it "generates the cli script" do
+      expect { subject.run %w[generate --quiet] }.to output_approval('cli/generate/quiet')
+      expect(File).to exist(cli_script)
+    end
+
+    context "when source files already exist" do
+      before do
+        expect { subject.run %w[generate --quiet] } #.to output_nothing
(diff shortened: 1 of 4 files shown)


spec/bashly/commands/generate_spec.rb
```diff
@@ -51,6 +51,34 @@
...
+    end
+
+    it "generates the cli script" do
+      expect { subject.run %w[generate --quiet] }.to output_approval('cli/generate/quiet')
+      expect(File).to exist(cli_script)
+    end
+
+    context "when source files already exist" do
+      before do
+        expect { subject.run %w[generate --quiet] } #.to output_nothing
```
wolfgang42, line 70: The rspec_approvals gem [doesn't seem to like empty output](https://github.com/DannyBen/rspec_approvals/blob/v0.9.2/lib/rspec_approvals/matchers/base.rb#L16), which is awkward in this situation. Since you're the creator of both gems I figured you'd have an idea about what to do here.

DannyBen:
Excellent idea, and a great implementation - thanks for both.

I will review the PR locally, and probably make a few adjustments, and fix whatever problem there is with `rspec_approval`.

> it does still print created #{file} after generating any new files).

Why is that? Shouldn't `--quiet` be completely quiet? Then it might bother someone else, who might want a `--completely-quiet` flag, no?

spec/bashly/commands/generate_spec.rb
DannyBen: Well, it is by design. An empty output can be tested with `.to be_empty` and there is no need for an approval matcher. So, it is usually a mistake when you use any of the approval matches, with an empty output. But don't worry about it, I will take a deeper look in the spec and fix accordingly.

DannyBen:
I have made my updated in #112, let us continue discussion there.

I will also probably merge #110 into it.

PR #110, opened 2021-09-30
Fix generate spec to actually test for overwriting

wolfgang42:
The bashly.yml that the spec uses does not define a "cli-get" command, so if the generator *were* overwriting this file the spec would not catch it. Change the spec to use a command name that the example yml does have, so that it would be overwritten if this happens.


DannyBen:
You are absolutely right, nice catch!

And thanks for the separate PR. I have merged this one into #112 as well.

PR #113, opened 2021-09-30
Make bashly-generated code `set -u`-safe

wolfgang42:
`set -u` ("no unset variables") is a useful shell option that catches a lot of bugs caused by things like misspelled variables. This PR updates all of the Bashly templates to use `set -u`-safe idioms so it can be set globally for bashly-managed scripts.

This is currently a DRAFT: there are some broken things (see PR comments) but filing it here for discussion.


examples/config-ini/src/lib/config.sh
```diff
@@ -17,11 +17,12 @@ config_init() {
   [[ -f "$CONFIG_FILE" ]] || touch "$CONFIG_FILE"
 }
 
-# Get a value from the config
+# Get a value from the config.
```
wolfgang42, line 20: This seems to have been an older version of this file, I copied from lib and all these changes showed up.

(This is a more general question I have about how bashly handles library changes given that it's just copying them in, but I think that's a separate issue.)

examples/environment-variables/src/verify_command.sh
```diff
@@ -4,6 +4,6 @@ echo "# you can edit it freely and regenerate (it will not be overwritten)"
 inspect_args
 
 echo "environment:"
-echo "- API_KEY=$API_KEY"
+echo "- API_KEY=${API_KEY:-}"
```
wolfgang42, line 7: As you can see I also had to make some changes to some of the examples to make them safe as well.

lib/bashly/views/script/header.erb
```diff
@@ -1,3 +1,8 @@
 #!/usr/bin/env bash
 # This script was generated by bashly (https://github.com/DannyBen/bashly)
 # Modifying it manually is not recommended
+
+if [[ "${BASH_VERSINFO:-0}" -lt 4 ]]; then
+  printf "<%= strings[:unsupported_bash_version] -%>\n"
+  exit 1
+fi
```
wolfgang42, line 8: Moved this to the top of the file so Bash 3 can see this and early exit before it gets to any syntax it doesn't understand (it was choking on trying to parse `[[ -v` IIRC). This also raises the question of if any of the other idioms I'm using require a newer version of Bash. Does bashly have a policy on minimum bash version? I briefly investigated running the tests on Bash 4.0 but couldn't immediately find a compiled copy of that which would actually run on my system.

lib/bashly/views/command/initialize.erb
```diff
@@ -2,12 +2,7 @@
 initialize() {
   version="<%= version %>"
   long_usage=''
-  set -e
-
-  if [[ "${BASH_VERSINFO:-0}" -lt 4 ]]; then
-    printf "<%= strings[:unsupported_bash_version] -%>\n"
-    exit 1
-  fi
+  set -eu -o pipefail
```
wolfgang42, line 5: **DANGER DANGER:** This will obviously break a lot of scripts if merged in. I stuck this here for testing but I'm not sure what the best way to handle this is. Maybe just adding this to every example's `initialize.sh`?

spec/bashly/commands/generate_spec.rb
```diff
@@ -64,7 +64,7 @@
       expect { subject.run %w[generate -w function] }.to output_approval('cli/generate/wrap-function')
       expect(File).to exist(cli_script)
       lines = File.readlines cli_script
-      expect(lines[5]).to eq "function() {\n"
+      expect(lines[10]).to eq "function() {\n"
```
wolfgang42, line 67: These offsets changed because the Bash 4 check pushed them down

spec/bashly/integration/bash_spec.rb
```diff
@@ -8,6 +8,7 @@
       command = "docker run --rm -v $PWD:/app bash:3 bash /app/download"
 
       Dir.chdir "examples/minimal" do
+        system "bashly generate 2>&1 >/dev/null"
```
wolfgang42, line 11: This test was using whichever version of the generated script was lying in the worktree, which was very confusing until I figured out what was happening.

DannyBen:
I am not so sure about the necessity of this.
There are other, less strict ways to catch bugs.

If users want it in their script, they can add it to the initialization sequence, and for bashly itself, I am not sure it is required.

Have you seen a bug that this would have helped catch?

examples/environment-variables/src/verify_command.sh
DannyBen: This is exactly one of the reasons I am not a fan of strict setup.
In some cases, it is perfectly legitimate to have empty, unset variables - it is a part of nature... :)

lib/bashly/views/script/header.erb
DannyBen: Hmm... no, I don't like this. My reasoning:

1. This type of check is exactly where it is supposed to be, logically (in my mind) - in the initialization sequence.
2. Of course, this requires that a) people who develop the bashly gem, havs bash 4, and b) that anything that appears before this check (which is just the 3 lines in the `initialize` function), are not going to fail in old bash versions.

To my knowledge, the current setup - from the standpoint of the generated script users - will always fail gracefully on bash 3. Am I wrong? If not, then I understand that this change is only meant to accommodate bash gem developers, and it is not something we should do, since having bash 4 when developing the bashly gem is a mandatory prerequisite in my mind.

On a more philosophic note - Bash 3 is over a decade long - we cannot encourage bad behavior by operating system vendors, no matter how powerful they are... 

> Does bashly have a policy on minimum bash version?

Yes, [here](https://bashly.dannyb.co/installation/#prerequisites) - bash 4 is a must.

spec/bashly/integration/bash_spec.rb
DannyBen: But why would the code in the example/minimal folder be any different, and require regeneration?

examples/environment-variables/src/verify_command.sh
wolfgang42: Yes, as a rule I do not want to force this on users. But it was useful to make all of the examples work in this mode, so that I could run the tests in this mode and easily make sure that all of the different ways of generating a script were still working.

lib/bashly/views/script/header.erb
wolfgang42: > the current setup - from the standpoint of the generated script users - will always fail gracefully on bash 3. Am I wrong?

Yes, this is not correct. If the `_command.sh` files from the user contain any syntax that Bash 3 does not understand, it will fail during parsing rather than gracefully with an explanatory error.

I will extract this change into a separate PR and make a test case for this situation, since it's confusing to have it mixed up with this PR.

spec/bashly/integration/bash_spec.rb
wolfgang42: It's not the code in the folder that is different, but rather the way the spec is using it. The `examples_spec.rb` file runs the examples by their `test.sh`, which all have `bashly generate` at the top of them and so regenerate the script before calling it. This spec is different: it calls the generated script directly (it has to, since `bashly` is not available in the `bash:3` container) and so unless we do so explicitly the script is not regenerated in this case.

wolfgang42:
> If users want it in their script, they can add it to the initialization sequence

This is the trouble. If you do that the generated script will crash because it runs `set -u` in initialize() and then immediately crashes while trying to parse arguments.

I do agree that this shouldn't be forced on users, hence my comment on `initialize.erb` about having added it there for testing (so I could run the specs and see what was crashing and needed to be fixed). I do think that all of the bashly tests should be run with this mode enabled, so that they can catch these problems.

> for bashly itself, I am not sure it is required

You mean to catch bugs in bashly? I agree, I don't think this is very important for checking bashly itself, but it is important to have so that user scripts can have this option set without having to turn it off before calling into bashly-generated code.

> Have you seen a bug that this would have helped catch?

Yes, this PR was prompted by me writing a bug that accidentally deleted a file because I made a mistake when writing the argument name. I then went to add `set -u` to my initializer to keep that from happening again and discovered that I couldn't do that because bashly was generating code that didn't work properly with this option set.

On a side note, I find it interesting that bashly sets `-e` automatically, especially in light of your comments about `-u`. Out of curiosity, what was the reason for setting that particular option?

DannyBen:
Ok, I see your point now. You wish to make the generated scripts "set -u compliant", and not add "set -u" as a default. With this I can agree. Just need to have a PR clean of anything that is debatable.

> Out of curiosity, what was the reason for setting that (set -e) particular option?

Setting `set -e` makes bash behave like other programming languages. When there is an error, it stops. This should be a default behavior in my view. Do you feel that it should not be in the initialize function?

lib/bashly/views/command/initialize.erb
DannyBen: We can generate this line conditionally, for example, if the environment variable `BASHLY_STRICT` is set. And then `ENV['BASHLY_STRICT'] = '1'` can be added to the spec helper.

DannyBen:
BTW - I was thinking that perhaps another way to tackle this problem it to add an optional `before` and `after` function placeholders. This will run before and after calling any user code. If this was available, you could have placed your `set -u` in the `src/before.sh` partial.

This will have this logical order:

1. `initialize` (user injectable)
2. *other bashly code*
3. `before` (user injectable)
4. *user code*
5. `after` (user injectable)

DannyBen:
By the way, I am still OK with continuing with this PR - if it only contains changes that make bashly `set -u` compatible, and nothing more (except of course the adjustments to the specs as needed).

If you want to clean it up, and mark it ready for review, I will merge.

wolfgang42:
Putting this PR on hold pending discusison on https://github.com/DannyBen/bashly/pull/114#issuecomment-939089510

wolfgang42:
That sounds like a reasonable approach; I'll try to make some time soon to clean up this PR to fix bash 3 support.

Also, continuing this discussion:
>> Out of curiosity, what was the reason for setting that (set -e) particular option?
>
> Setting `set -e` makes bash behave like other programming languages. When there is an error, it stops. This should be a default behavior in my view. Do you feel that it should not be in the initialize function?

I don't have a strong opinion either way, I just feel like this may be an inconsistency. `set -u` also makes bash behave more like other programming languages (most of which require declaring or at least setting a variable before using it), so I was wondering why Bashly would turn on one but not the other.

DannyBen:
> I just feel like this may be an inconsistency. set -u also makes bash behave more like other programming languages.

Well - bashly is just a tad opinionated.
I adapt the Ruby moto - make developers happy. Making something more strict, usually doesn't.

Also, don't forget that bash was designed to allow these unset vars - as can be seed in this very PR - natural ways of writing things in bash, fail with this enabled.

wolfgang42:
Per your note on https://github.com/DannyBen/bashly/pull/121#issuecomment-940700288, LMK what you decide about this PR. Happy to make it Bash 3-compatible as we discussed.

DannyBen:
@wolfgang42 - I see you are no longer available to continue this, and since the master branch has diverged too much, I could not resolve its conflicts.

I am closing this in favor of #148, which I had to recreate manually based on your code.

Thanks for your help.

PR #114, opened 2021-10-08
Move bash 3 version check to top of file for early exit

wolfgang42:
Bash executes the script as it's reading it. When it encounters a function, it parses the function body before continuing. The bashly-generated script has *all* of its code in functions which appear before any code is executed (the initialize/run functions are only called at the bottom of the script). Normally this is useful, since it ensures that bash has loaded/parsed the whole script before beginning execution; however, in the case of this version check it means that if any of the functions contain syntax that Bash 3 doesn't understand, it will choke on that and crash (with a much less helpful error) before it has a chance to execute the version check:

    /app/cli: line 117: conditional binary operator expected
    /app/cli: line 117: syntax error near `missingno'
    /app/cli: line 117: `  if [[ -v missingno ]]; then'

This PR adds a test for this situation and moves that check to the very top of the file, where it will be executed and trigger an early exit before bash tries to parse the rest of the script, so that it will work correctly regardless of the remainder of the contents.

(Extracted from #113 for clarity.)


DannyBen:
> Yes, if you check out only the first commit of the series you will find a failing test. (Though I made it a `test_command` rather than `root_command`, happy to rearrange if you prefer the former.)

no need to rearrange - just a reference.

DannyBen:
Well.

I am able to reproduce the issue clear as day.

Did you consider the `bashly generate --wrap my_function` use case?
As far as I understand it, this fix effectively kills it.

DannyBen:
Yeah. Well, my current position is this:

I realize you have put a lot of thought and effort into this, and I appreciate it.
However, as of this very moment, I feel that #116 is a more appropriate solution.

My reasoning:

1. Bash 3 is not a priority for bashly. Old software belongs on a Wikipedia page, not making the developer's life miserable. 
2. I suspect that most bashly users develop scripts for themselves, their friends or colleagues. In that case, they control the environment. In any other case, it will be their responsibility to test and fix their own script on bash 3, if they choose to do so.
3. Out of the box scripts generated by `bashly g` will fail gracefully on bash 3 - anything beyond that, is developer land, and their responsibility.
4. In case other developers - like you - feel that they must add such a blocker at the very top of the script, they will be able to do so with #116.

Your thoughts?

wolfgang42:
My thoughts on your reasoning, as requested:

> 1. Bash 3 is not a priority for bashly. Old software belongs on a Wikipedia page, not making the developer's life miserable.

I wish!

> 2. I suspect that most bashly users develop scripts for themselves, their friends or colleagues. In that case, they control the environment. In any other case, it will be their responsibility to test and fix their own script on bash 3, if they choose to do so.

This is true, though less inscrutable error messages are still helpful if their friend or colleague is using a Mac.

> 3. Out of the box scripts generated by bashly g will fail gracefully on bash 3 - anything beyond that, is developer land, and their responsibility.

This is true, but the out-of-the-box scripts also don’t actually *do* anything. It would be nice if, when you actually *use* bashly, the check that it ostensibly provides didn't break because of a seemingly innocent implementation.

<hr/>

On a much more general note, a quick summary of where we stand:

* The work on #113 (`set -u` safety) depends on this PR (Bash 3 early exit); otherwise *all* bashly scripts will fail ungracefully in Bash 3.
* You would prefer not to merge this PR (resulting in keeping the status quo, that the bashly framework is “Bash 3 clean” and if individual script authors use syntax that isn’t it’s up to them to know about that),
* Although 113 on its own is fine (if it’s cleaned up to not interfere with *everyone's* scripts, which makes sense) it would break the “Bash 3 cleanness” (there may be a workaround to this, but IMO it's not worth my effort) and can’t be merged either.
* Instead, you have proposed using a hooks system (#116) so that individual scripts can implement these checks (and others) as needed. 

My philosophy is that error handling and correctness checks should be pushed as far upstream as possible, to make it easier for more people to benefit from them; but since the hooks will provide a workaround that fits my needs I’m OK with dropping my two open PRs and using hooks instead. Yours is also a reasonable stance, just a different one; and as it's not my project I leave it up to you to make the final decision on which direction to go in.

**TLDR:** Feel free to close #113 and #114 if you don't like them; I’d like to see Bashly have better support for these situations but am OK with keeping my opinions confined to my own scripts.

<hr/>

> (and for future reference, to avoid work that may not be merged, perhaps lets open new ideas in an issue before PR).

(This is a tricky balancing act. In the case of these two PRs, most of what we’ve been discussing has come up as a result of having code to look at and wouldn't necessarily have been obvious before I started implementing. This is why I submitted #114 as a draft needing cleanup, once I realized that it wasn’t as simple as I had anticipated and would need discussion before merging.)

DannyBen:
And BTW - I may still decide to merge this. Perhaps after some sleep I will look at it differently.
Just a little concerned about the `--wrap function`.

DannyBen:
I have reviewed all changes, looks good. Now only left to decide if we want this or not.

It is now somewhat easier to accept, since there is an ability for the user to replace it with a custom header. It is the same reason that renders it a little less necessary though. Just a matter of what we feel is a better default, and I know your stand on the subject is that *this* is a better default.

If we merge it, I will also probably want to change the header for when generating with `--wrap`.

For now, lets keep it open until I decide one way or another, or until more evidence comes to light to tip the scale.

PR #119, opened 2021-10-09
Remove generated example executables from the repo

### LINKED ISSUE #118: Remove generated examples from the repositories
Any code change in bashly that alters the generated code even a little bit causes all generated example executables to be changed, making PR review much more messy than it should be with dozens of changed example files.

I believe we can safely remove them and `.gitignore` them.

DannyBen:
Closes #118


wolfgang42:
@DannyBen merged in master on both PRs

PR #121, opened 2021-10-10
Finalize #114

DannyBen:
This PR supersedes (and includes) #114.

Changes on top of the original:

- Bash 3 bouncer is now in its separate partial (but is still considered a "header", in case it is replaced by the user).
- Bash 3 bouncer will only be included when a standalone script is generated. When a function is generated, it is left to the user to take care of it some other way.
- Some high churn specs were modified to use approvals instead of exact text matches


DannyBen:
@wolfgang42 - this is mostly your PR, with some flavor.

I will leave it here to cool off for a day or two, to give you a chance to take a look if you want, and will merge afterwards.

PR #122, opened 2021-10-13
Allow empty string as flag argument

wolfgang42:
(This overlaps with the changes in #113, but came up in an unrelated context and seemed worthy of discussion on its own merits.)

I was explicitly passing in the empty string as an argument to a flag, but bashly treated this as though the argument was absent, which is not the behavior I wanted. This PR changes that so that it will accept the empty string correctly, and adds a spec using the example of a blank password.

I think the same change should also be made to non-flag arguments, though I haven't investigated that case yet.


DannyBen:
Lets simplify the argument.

From the bashly manifesto:

> Bashly is responsible for ... Preventing your script from running unless the command line is valid.

More often than not, for most users, with most flags - an empty string is an invalid input.
Making this change, will force people to have to write custom error handling in every function they have. This is unacceptable.

Although I am yet to see a use case that does not have a better solution, I can accept that it is a matter of style.

How about enabling these empty params conditioanlly, as follows:

- Allowing both arg params and positional params to accept empty string (using the syntax proposed in this PR)
- Adding a check after the param was captured, and **aborting** if it is empty with a new text message that is more appropriate.
- This will allow to pass if the arg has an `allow_empty: true` in their config

How do you feel about that?

wolfgang42:
I was actually thinking more about this last night and was going to suggest `allow_empty` as a possibility. (Actually, I had the idea for a more general validations system, which I have written down in #125 for discussion).

Your plan sounds good, except—I am not sure about making empty strings allowed and then later adding the `allow_empty` option again. It seems confusing for script writers to have this validation go away and come back again some time later, though I suppose since bashly is still v0.x this is maybe to be expected.

The extra test line was mainly to demonstrate the point, for discussion; I can definitely make a proper fixture test for both flags and arguments, if we decide that this PR is a good approach.

Incidentally, I just want to say: I have sometimes taken a somewhat combative tone in these discussions, because I have quite strong opinions about how some of these things should work, but I appreciate the way that you also have similarly strong opinions and a cohesive vision for how Bashly should work. I feel that, though it may take a while to get there, the final result will be the stronger for it.

DannyBen:
> I take some time soon to clean this one up per your plan #122 (comment) above so it can be merged.

Sounds great.

cjbrigato:
The empty argument is, in niche case, the only way to respect POSIX guidelines on utility argument syntaxes, guideline that shaped much of our aquired tastes up to now.

> Guideline 7:
> Option-arguments should not be optional.

An empty argument permit fullfiling this rule. Later validation or de-facto broken usage is up to deal with it.

wolfgang42:
@DannyBen I've created a fixture test and removed the example test I had added. It turns out that positional arguments already worked, but I have added a test case for them anyway.

PR #130, opened 2021-10-17
Add optional arg/flag validation functions

### LINKED ISSUE #125: Idea: argument validation system
I was thinking last night about the argument validation we've been discussing in #122. I had the idea of an `allow_empty_string` config setting, but then I thought—why should the empty string be special? There are a lot of different validations that might be important for an argument, such as whether the file exists or if the value is an integer.

To that end, here's a straw-man proposal for a validations system:

* Flags and arguments have a new config key, `validate`, which takes an array of validations to check the value for. (Optional: If there is only one validation, it can be passed as a bare string.)
* Validations are defined by a function named `validate_${validation_name}` which takes one argument, the value to be validated, and echos a string if the value is not valid.
* Bashly has some built-in validations, but additional ones can be defined in `initialize.sh`

Here are some example validations which people might write:
```sh
validate_not_empty() {
  [[ -z "$1" ]] && echo "must not be empty"
}
validate_file_exists() {
  [[ -e "$1" ]] && echo "must be an existing file"
}
validate_commit_exists() {
  git rev-parse --quiet --verify "$1^{commit}" || echo "must be a valid git commit"
}
validate_is_integer() {
  [[ "$1" =~ ^[0-9]+$ ]] || echo "must be an integer"
}
```

And an example config:
```yaml
commands:
- name: upload
  args:
  - name: source
    required: true
    validate: file_exists
  flags:
  - long: --scratch-file
    validate: [file_readable, file_writable]
```

WDYT?

DannyBen:
This PR adds a way to easily validate arguments and flag-arguments using either built in validations, or custom validation functions.

### Checklist

- [x] Feature foundations
- [x] Current specs should pass
- [x] Feature specs
- [x] Feature examples
- [x] Initial set of built in validations in `bashly add validations`

### Notes

- Argument validations had to be added in two places: required arg filter, and the regular arg filter.
- Validation functions must start with `validate_` and must return either an error string (suffix) or an empty string.
- The error message will always start with `validation error:`, and can be changed in the strings config.

Closes #125

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -43,3 +43,4 @@ Each of these examples demonstrates one aspect or feature of bashly.
 - [colors](colors#readme) - using the color print feature
 - [yaml](yaml#readme) - using the YAML reading functions
 - [completions](completions#readme) - adding bash completion functionality
+- [validations](validations#readme) - adding argument validation functions
--- a/examples/validations/README.md
+++ b/examples/validations/README.md
@@ -64,7 +64,61 @@ commands:
 ### `$ ./calc -h`
 
 ```shell
-sh: 1: ./calc: not found
+calc - Sample application demonstrating validations
+
+Usage:
+  calc [command]
+  calc [command] --help | -h
+  calc --version | -v
+
+Commands:
+  add   Add two numbers
+
+Options:
+  --help, -h
+    Show this help
+
+  --version, -v
+    Show version number
+
+
+
+```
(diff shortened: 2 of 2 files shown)


DannyBen:
This can be merged, there is only one built in `integer` validation for now, but I feel additional validations can be added post-merge, as this has become a large enough PR.

PR #135, opened 2021-10-22
Allow empty string as argument (replacing #122)

DannyBen:
Attempt resolution for, and closes #122 

**[Reference: How to check if a variable is set in Bash?](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash)**


wolfgang42:
LGTM, thanks for dealing with the merge conflict and additional tests.

PR #148, opened 2021-10-30
Make generated scripts compatible with bash strict mode

DannyBen:
- [x] Make all the generated bash code compatible with bash strict mode (`set -euo pipefail`).
- [x] Add an environment variable `BASHLY_STRICT`, that if set, will enable strict mode in the script's initialization function
- [x] Set the specs to set `BASHLY_STRICT=1` so that from this point on, all new bash code in bashly will also be compatible.

Closes #113

### DIFF
--- a/lib/bashly/views/command/run.erb
+++ b/lib/bashly/views/command/run.erb
@@ -9,7 +9,7 @@ run() {
   <%- condition = "if" -%>
   <%- deep_commands.each do |command| -%>
   <%= condition %> [[ $action == "<%= command.action_name %>" ]]; then
-    if [[ ${args[--help]} ]]; then
+    if [[ -v args[--help] ]]; then
       long_usage=yes
       <%= command.function_name %>_usage
     else
--- a/lib/bashly/views/command/run.erb
+++ b/lib/bashly/views/command/run.erb
@@ -1,15 +1,15 @@
 # :command.run
 run() {
-  declare -A args
-  declare -a other_args
-  declare -a input
+  declare -A args=()
+  declare -a other_args=()
+  declare -a input=()
   normalize_input "$@"
   parse_requirements "${input[@]}"
 
   <%- condition = "if" -%>
   <%- deep_commands.each do |command| -%>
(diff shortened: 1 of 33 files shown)


DannyBen:
@wolfgang42 - this is the new strict mode support. I will let it sit here for a day or two in case you are around and wish to review before I merge.

wolfgang42:
Thanks for doing the work to port all these changes onto the latest version. Really all I did to write my PR was to `set -u` and then keep fixing things until all the tests pass, so since they are I don't think I have anything else to add.

I was somewhat confused about how you didn't trigger the shellcheck SC2102 bug with `[[ -v`, but on investigation I think that was caused by my fix for [this line in #135](https://github.com/DannyBen/bashly/pull/135/files#diff-9ea5cb9d1d6f5a6cded1b564e4a5a2d89c52af3bdd11edf05a9659d38e88260aL3-R3) which you happened to resolve in a different way that doesn't trigger the bug.

DannyBen:
Yes, that line was removed in #139

DannyBen:
One thing I just realized, is that the `[[ -v varname ]]` syntax is only supported in bash 4.2 onward. I wonder if there is a more compatible way to solve the small number of checks that had to resort to this syntax.

I will try setting the array to something.

[Reference SO question](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash)

wolfgang42:
Ah, good catch. I believe you should be able to use `[[ -n ${var+x} ]]` universally in lieu of `[[ -v var ]]`.

lib/bashly/views/command/run.erb
```diff
@@ -1,15 +1,15 @@
 # :command.run
 run() {
-  declare -A args
-  declare -a other_args
-  declare -a input
+  declare -A args=()
+  declare -a other_args=()
+  declare -a input=()
   normalize_input "$@"
```
DannyBen, line 6: I was able to get rid of the `[[ -v arg ]]` test and the unbound error they solved, by assigning an empty array.

PR #152, opened 2021-11-11
Add support for private commands

### LINKED ISSUE #149: Feature request: allow sub-commands to be hidden from help text
Hi, @DannyBen! :wave: 

We are happy users of Bashly at @kolonialno, having written much of our internal data science tooling with it, thanks a lot! :bow: Let me start by explaining what I want to _achieve_ in order to prevent us having an [XY Problem](https://xyproblem.info/) on our hands.

## Problem to be solved

We are starting to amass a lot of sub-commands in on of our internal CLIs, and would like to clean up the help text and logic a bit. For instance, one of our commands `cli_name sub_command <arg>` is in practice just some glorified switch logic which invokes either `cli_name sub_command_arg1`, `cli_name sub_command_arg2`, or `cli_name sub_command_arg3` based on the value of `<arg>`. In fact, we intend our internal users to never invoke `cli_name sub_command_arg{1,2,3}` directly, and only invoke it through `cli_name sub_command`. We could in theory move the logic of the sub-commands into library functions, but then we lose all the nice tooling provided by Bashly for those commands; argument parsing and validation, and dependency checking come to mind.

## Proposed solution

If bashly provided an additional optional key in the `command` object configuration, let's say `hidden: true`, which excludes the given sub-command from the output of `cli_name` and `cli_name --help`, but still allows the command to be directly executed with `cli_name sub_command` and documentation with `cli_name sub_command --help`.

Another use-case for this parameter would be to release sub-commands in "Beta mode", hidden from most users, but you can do a soft launch by letting a subset of the CLI users know that the new sub-command is available on direct invocation.

## Alternative solution

Another solution that comes to mind is something more along the lines of "library functions" only for commands instead. Perhaps a command could be annotated as a "library command" or "private command", making it exclusively available in the namespace of the command implementations, but not outside from the perspective of the CLI user.

Thanks in advance,
Jakob

DannyBen:
- [x] Implementation
- [x] Specs / Example
- [x] Docs

Closes #149

### DIFF
--- a/examples/colors/README.md
+++ b/examples/colors/README.md
@@ -66,5 +66,18 @@ echo
 
 ```
 
+### `$ NO_COLOR=1 ./colorly`
+
+```shell
+Message Recevied:
+
+  => hello colors
+  ==> hello colors
+  ===> hello colors
+
+
+
+```
+
 
 
--- a/examples/command-private/.gitignore
+++ b/examples/command-private/.gitignore
@@ -1,0 +1,6 @@
+src/connect_ftp_command.sh
+src/connect_ssh_command.sh
+src/download_command.sh
+src/initialize.sh
+src/upload_command.sh
+cli
--- a/examples/command-private/README.md
+++ b/examples/command-private/README.md
@@ -13,3 +13,129 @@ $ bashly generate
 <!-- include: src/connect_command.sh -->
 
 -----
+
+## `bashly.yml`
+
+```yaml
+name: cli
+help: Sample application with private commands
+version: 0.1.0
+
+commands:
+- name: connect
+  short: c
+  help: Connect to the metaverse
+
+  args:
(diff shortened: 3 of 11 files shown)


wolfgang42:
That was quick! Just tried it out and it looks like exactly what I wanted, my help message is much less cluttered now.

DannyBen:
Alright - cool. 
I will implement specs/example then, and give 24 hours for the OP to comment.
I will cut a version tomorrow unless there will be objections/requests from anyone.

JakobGM:
Sorry for the late reply, thank you so much, this looks good :relaxed:

DannyBen:
Great. New release will happen sometime in the next few days, possibly tomorrow.
Just waiting for some decisions on the two open PRs.

PR #154, opened 2021-11-12
Add ability to import external snippets in bashly.yml

### LINKED ISSUE #153: Allow splitting `bashly.yml` to smaller files (+ support front matter)
As discussed in #151

DannyBen:
### Checklist

- [x] Implement ability to load YAML snippets from bashly.yml
- [x] Remove front matter from user files before merging them in
- [x] Error gracefully and informatively when the imported file is not found
  - [x] ... or when it is not a valid YAML (for front matter)
- [x] Specs
- [x] Example
- [x] Docs

### Key Changes

- Added `ComposeRefinements` module, which applies to hashes and arrays, and imports YAML files when it encounters a hash with `"import" => "path"`
- The refinement is used in the generate command, when loading the user config.
- Front matter is supported natively by `YAML::load_file`.
- Added `String#remove_front_matter` extension
- The extension is used in one place only - when loading user files.

Closes #153

### DIFF
--- a/examples/colors/README.md
+++ b/examples/colors/README.md
@@ -66,5 +66,18 @@ echo
 
 ```
 
+### `$ NO_COLOR=1 ./colorly`
+
+```shell
+Message Recevied:
+
+  => hello colors
+  ==> hello colors
+  ===> hello colors
+
+
+
+```
+
 
 
--- a/lib/bashly/commands/generate.rb
+++ b/lib/bashly/commands/generate.rb
@@ -1,6 +1,8 @@
 module Bashly
   module Commands
     class Generate < Base
+      using ComposeRefinements
+
       help "Generate the bash script and required files"
 
       usage "bashly generate [--force --quiet --upgrade --wrap FUNCTION]"
@@ -116,7 +118,7 @@ module Bashly
       end
 
       def config
-        @config ||= Config.new "#{Settings.source_dir}/bashly.yml"
+        @config ||= Config.new("#{Settings.source_dir}/bashly.yml").compose
       end
 
       def command
--- a/lib/bashly/extensions/string.rb
+++ b/lib/bashly/extensions/string.rb
@@ -27,4 +27,8 @@ class String
     gsub(/\s+\n/m, "\n\n").lines.reject { |l| l =~ /^\s*##/ }.join ""
   end
 
+  def remove_front_matter
+    split(/^---\s*/).last
+  end
+
 end
--- a/lib/bashly/refinements/compose_refinements.rb
+++ b/lib/bashly/refinements/compose_refinements.rb
@@ -1,0 +1,36 @@
+require 'yaml'
+
+module ComposeRefinements
+  refine Hash do
+    def compose(keyword = 'load')
+      result = {}
+      each do |k, v|
+        if k.to_s == keyword
+          sub = YAML.load_file(v).compose keyword
+          if sub.is_a? Array
+            result = sub
+          else
+            result.merge! sub
+          end
+        elsif v.respond_to? :compose
+          result[k] = v.compose keyword
+        else
+          result[k] = v
+        end
+      end
+      result
+    end
+  end
+
+  refine Array do
+    def compose(keyword = 'load')
+      map do |x|
+        if x.respond_to? :compose
+          x.compose keyword
+        else
+          x
+        end
+      end
+    end
+  end
+end
--- a/lib/bashly/refinements/compose_refinements.rb
+++ b/lib/bashly/refinements/compose_refinements.rb
@@ -24,10 +24,10 @@ module ComposeRefinements
     def safe_load_yaml(path)
       loaded = YAML.load_file path
       return loaded if loaded.is_a? Array or loaded.is_a? Hash
-      raise Bashly::ConfigurationError, "Cannot find a valid YAML in '#{path}'"
+      raise Bashly::ConfigurationError, "Cannot find a valid YAML in !txtgrn!#{path}"
 
     rescue Errno::ENOENT
-      raise Bashly::ConfigurationError, "Cannot find import file '#{path}'"
+      raise Bashly::ConfigurationError, "Cannot find import file !txtgrn!#{path}"
     end
   end
 
--- a/lib/bashly/refinements/compose_refinements.rb
+++ b/lib/bashly/refinements/compose_refinements.rb
@@ -6,7 +6,7 @@ module ComposeRefinements
       result = {}
       each do |k, v|
         if k.to_s == keyword
-          sub = YAML.load_file(v).compose keyword
+          sub = safe_load_yaml(v).compose keyword
           if sub.is_a? Array
             result = sub
           else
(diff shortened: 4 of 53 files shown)


examples/colors/README.md
```diff
@@ -66,5 +66,18 @@ echo
...
+### `$ NO_COLOR=1 ./colorly`
+
+```shell
+Message Recevied:
+
+  => hello colors
+  ==> hello colors
+  ===> hello colors
+
+
```
DannyBen, line 78: This and a couple of others were regenerated as a side effect, ignore.

lib/bashly/commands/generate.rb
```diff
@@ -116,7 +118,7 @@ def master_script_path
       end
 
       def config
-        @config ||= Config.new "#{Settings.source_dir}/bashly.yml"
+        @config ||= Config.new("#{Settings.source_dir}/bashly.yml").compose
```
DannyBen, line 121: This is where the composition happens, by just calling the `#compose` on the refined Hash, it will convert any `import: path` to the contents of the YAML represented by the path.

lib/bashly/extensions/string.rb
```diff
@@ -27,4 +27,8 @@ def lint
     gsub(/\s+\n/m, "\n\n").lines.reject { |l| l =~ /^\s*##/ }.join ""
   end
 
+  def remove_front_matter
+    split(/^---\s*/).last
+  end
```
DannyBen, line 32: This is called in one place only, when loading user content.

lib/bashly/refinements/compose_refinements.rb
```diff
@@ -0,0 +1,36 @@
...
+      map do |x|
+        if x.respond_to? :compose
+          x.compose keyword
+        else
+          x
+        end
+      end
+    end
+  end
+end
```
DannyBen, line 36: This file is the "heavy lifting". It adds refined functionality to hashes, and arrays that contain arrays or hashes, to allow `import: yaml_path`.

DannyBen:
@wolfgang42 this is your feature request - do you want to review?

DannyBen:
Docs added here: https://github.com/DannyBen/bashly-book/commit/7261c914d0334a6744e2355e643062b49eb4d1be?short_path=6519d7b#diff-6519d7b1f43d642f6713f9c087702a798815b1c19a73c354201f497053b5daac

wolfgang42:
Skimmed the docs briefly and it looks very promising. I'll try it out over the weekend, but I know you prefer a faster pace of development so don't let me stop you if you'd prefer to merge sooner.

DannyBen:
Excellent. Its a big enough feature, I would rather have your second opinion before merging.

wolfgang42:
Just tried it out; it works the way I was hoping and I didn't run into any technical issues.

However, I'm not sure about the design of the front matter. Bashly understands what's going on with YAML at the top of a shell script, but nothing else does, so it seems like there's great potential for problems here. Even for a human reader it's not obvious without knowing what's going on, and syntax highlighting definitely also gets in the way. (My editor picked up on an apostrophe in a YAML string and carried that through to the end of the file, making everything the wrong colors until I adjusted the quoting to make it happy.)

I think it should probably be formatted as a comment, or maybe a specially marked HEREDOC or something, to keep it still as valid shell syntax—though I recognize that this will make it no longer plain YAML, and so will require special handling, I think that's the lesser of two evils in this case.

DannyBen:
> Bashly understands what's going on with YAML at the top of a shell script, but nothing else does

Nothing else needs to.... except maybe a text editor.

> Even for a human reader it's not obvious without knowing what's going on

You can't say this categorically.
If people are familiar with front matter, it is crystal clear to them.
This feature is classified under advanced use. Most people are encouraged NOT to us it this way.

Front matter is a standard practice. It is *the* way to embed configuration in another document.

You can't complain about syntax highlighting on one hand, and then propose it to be as a comment on the other... 

Personally - I think that the configuration should stay in the YAML, and the shall code in the shell script. But if someone insists on having them together, then I think a YAML front matter is the most natural and native way of doing it.

I do not want to add any new inventions and hacks in the form of comments or specially marked anything.

wolfgang42:
>> Bashly understands what's going on with YAML at the top of a shell script, but nothing else does
>
> Nothing else needs to.... except maybe a text editor.

Well, text editors are a pretty big “something else.” But also the author might be using `shellcheck`, which complains that `---` is some kind of misplaced flag. Or `shfmt`, which will mangle the indentation of the YAML. Or they want to source the individual command scripts for a testing framework or something. Or another use case I haven't thought of entirely.

On the whole, I really think that the contents of a file named `.sh` ought to be parseable as a shell script, if at all possible.

>> Even for a human reader it's not obvious without knowing what's going on
>
>You can't say this categorically.
> If people are familiar with front matter, it is crystal clear to them.

I didn't mean to say it categorically, just that YAML is probably not the first thing most people would expect to see at the top of a file labeled as a shell script. Of course front matter is not uncommon, and I expect that most readers would figure out what they were looking at pretty quickly, but it is disconcerting.

> You can't complain about syntax highlighting on one hand, and then propose it to be as a comment on the other...

Er, why not? My complaint about the syntax highlighting was that it's flat-out *wrong*, putting the YAML into a comment would fix that. (The YAML would still not be highlighted, of course, but at least the editor won't be trying any more.)

> Personally - I think that the configuration should stay in the YAML, and the shall code in the shell script. But if someone insists on having them together, then I think a YAML front matter is the most natural and native way of doing it.
>
> I do not want to add any new inventions and hacks in the form of comments or specially marked anything.

I can see your reasoning, and generally I think that following common standards like this would be a good thing. In this situation, though, I disagree that the front matter is most intuitive. To me it seems like *more* of a hack—perhaps not a hack in Bashly, but in all of the surrounding complications.

Put it this way: though I'm the one that proposed this originally, I don't think I'll use it in it's current state; I think the downsides I've discussed above outweigh the convenience of having this built in. It seems to me like it needs more thought before it's ready to be added as a feature.

DannyBen:
> I really think that the contents of a file named .sh ought to be parseable as a shell script, if at all possible.

Front matter existence contradicts this statement. It only ever exists in foreign files.

> But also the author might be using shellcheck, which complains that --- is some kind of misplaced flag. Or shfmt

Why run tools on the partials, run them on the final.

> YAML is probably not the first thing most people would expect to see at the top of a file labeled as a shell script

Who is *people*? The only people looking at it are probably the same people who put it there....

---

I am currently leaning towards merging it, but I will wait for a few more days perhaps some improvement pops to mind.

DannyBen:
> Er, why not? My complaint about the syntax highlighting was that it's flat-out wrong

I thought you meant the fact that there is no syntax highlighting on the YAML, not that it scrambles the syntax highlighting of the shell script.

The only case where I see it scrambling the shell syntax highlighting is when you have an unclosed quote, which is super easily fixable:

```yaml
hello: "User's folder"
---
echo "# this file is located in 'src/upload_command.sh'"
echo "# code for 'cli upload' goes here"
echo "# you can edit it freely and regenerate (it will not be overwritten)"
inspect_args
```

PR #157, opened 2021-11-14
Add basic validation for key config properties

### LINKED ISSUE #156: Unhelpful error "undefined method `size' for nil:NilClass" with invalid bashly.yml
This YAML (note the misspelled key):
```yaml
name: example
commands:
  - mane: test # [sic]
```

produces the following output:
```
creating user files in src
skipped src/initialize.sh (exists)
skipped src/_command.sh (exists)
NoMethodError
undefined method `size' for nil:NilClass
```

and also creates an extraneous file called `src/_command.sh`. 

It took me a while to figure out where things had gone wrong, seems like there is should be some added validations somewhere to catch the case of an invalid command object?

DannyBen:
Closes #156

### DIFF
--- a/lib/bashly/script/argument.rb
+++ b/lib/bashly/script/argument.rb
@@ -4,6 +4,10 @@ module Bashly
       def usage_string
         required ? name.upcase : "[#{name.upcase}]"
       end
+
+      def verify
+        raise ConfigurationError, "Argument must have a name" unless name
+      end
     end
   end
 end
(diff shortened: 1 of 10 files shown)


DannyBen:
@wolfgang42 - as it turns out, all the models in the `Script` namespace already had a `verify` method called on initialize. It was largely unused.

I have added some very minimalistic tests to all models - mostly checking they have a name, which is key.

I think this is good enough to prevent unfriendly errors, but for a more advanced tests, perhaps a `bashly validate` command is more appropriate. Not sure it is needed for now, but if we believe it is, we can add an issue to track it.

PR #159, opened 2021-11-15
Validate the entire config prior to generating

### LINKED ISSUE #158: Consider adding `bashly validate`
to validate the config in its entirety.

This is in addition to the fix in #157, and following #156.

DannyBen:
- Improve validation of config file
- Run validation prior to running `bashly generate`
- Add `bashly validate` command to just run validation

Validation errors will show the full data-structure path, like so:

```
root.commands[0].commands[1].args[1].name must be a string
```

Closes #158

### DIFF
--- a/lib/bashly/cli.rb
+++ b/lib/bashly/cli.rb
@@ -10,6 +10,7 @@ module Bashly
 
       runner.route 'init',      to: Commands::Init
       runner.route 'preview',   to: Commands::Preview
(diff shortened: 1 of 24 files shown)


DannyBen:
@wolfgang42 - this is a more robust validation that validates that all provided types are the expected ones.

It does *not* complain about keys that are not expected. I chose to ignore this for now.

Thoughts?

PR #171, opened 2022-02-01
Allow specifying filenames for command partials

### LINKED ISSUE #170: Allow commands to have their own location
When using bashly for larger cli projects, the src folder can be a mess. 

For now I'm using more than 50 different commands. Some with splitted configs, but all SH files are located in the `src` directory. 

It would be cool if there is an option to provide a custom path per command, so I can handle my own script structure.

After diggin' the source code and docs, I couldn't find any option for that.

DannyBen:
Closes #170

- [x] Implementation
- [x] Specs
- [x] Examples
- [x] Docs


DannyBen:
Docs are ready in https://github.com/DannyBen/bashly-book/pull/6

Issue #216, opened 2022-04-28
feature request: enable herdocs by indenting with tabs instead of spaces


hojerst:
Problem: Bashly currently wraps command script in functions and indents the code. This effectivily disables the possibility to use heredoc's. The documented suggestion is to use a script in the lib folder as this is not indented. However, this breaks again when running bashly with `-w` flag as everything is wrapped in a function again.

Suggestion: Use tabs to indent functions instead of spaces. This allows the use of `<<-EOF` heredocs which CAN be indented, but only work when text is indented with tabs instead of spaces.

DannyBen:
How about running `unexpand -t2 your_final_script > your_tabbed_script`?

hojerst:
As I understood it, the generated script isn't really needed for 'human consumption' but rather a artifact. Do you disagree? I mean it is beneficial to have something readable here, which consistent tab usage won't remove, nontheless this is just generated and not meant for direct editiing.

If you strongly disagree with the default indention to be tabs: how about adding a option to use tabs as a argument to the `generate` command or at least documenting your workaround above. Currently there is NO option to use heredocs when you want to use the `-w` option of bashly. This hurts much more from a readability/usability standpoint than having a slightly less readable GENERATED script.

DannyBen:
> As I understood it, the generated script isn't really needed for 'human consumption' but rather a artifact. Do you disagree?

No. I agree.

> how about adding a option to use tabs as a argument to the `generate` command

I don't see it solving anything. 
As I see it, there are two ways to ensure the end result uses tabs instead of spaces:

1. Rewrite all of bashly's internal bash templates, to use tabs AND forcing the user to write THEIR code using tabs.
2. Post process, and just replace spaces with tabs in the generated script - which as I mentioned, can be done with one simple linux command.

> Currently there is NO option to use heredocs when you want to use the `-w` option of bashly.

How come? What is wrong with running `unexpand` on your generated script?

hojerst:
> > how about adding a option to use tabs as a argument to the `generate` command
> 
> I don't see it solving anything. As I see it, there are two ways to ensure the end result uses tabs instead of spaces:
> 
>     1. Rewrite all of bashly's internal bash templates, to use tabs AND forcing the user to write THEIR code using tabs.

I disagree. If bashly would indent e.g. the generated `*_command` functions with a single tab instead of 2 spaces, the user could use `<<-EOF` heredocs without a problem and still use spaces for normal indention:

e.g. src/test.sh
```shell
[...]

PR #217, opened 2022-04-28
Add support for generating tab-indented scripts using `BASHLY_TAB_INDENT`

DannyBen:
As discussed in #216

### DIFF
--- a/examples/heredoc/README.md
+++ b/examples/heredoc/README.md
@@ -1,24 +1,24 @@
 # Heredoc Example
 
-This example shows how (and where) you can use *Here document* (heredoc)
-constructs.
+This example shows how you can use *Here document* (heredoc) constructs.
 
-Since bashly applies indentation to most of your code, it is not possible to 
-use heredoc everywhere. Code in the `src/lib` directory is injected to your
-script as is, so in cases where you must use heredocs, this is where you can.
+By default, bashly indents all code using spaces. In order for heredoc to
+properly work in bash, all leading spaces within the heredoc document must use
+the tab character.
+
+Setting the environment variable `BASHLY_TAB_INDENT` to any value will make
+bashly replace the space indentation in the result, with tabs.
 
 This example was generated with:
 
 ```bash
 $ bashly init
-$ bashly add lib
 # ... now edit src/bashly.yml to match the example ...
 # ... now edit src/root_command.sh to match the example ...
-# ... now edit src/lib/heredocs.sh to match the example ...
-$ bashly generate
+$ BASHLY_TAB_INDENT=1 bashly generate
 ```
 
-<!-- include: src/root_command.sh src/lib/heredocs.sh -->
+<!-- include: src/root_command.sh -->
 
 -----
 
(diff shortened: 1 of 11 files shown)


DannyBen:
Thanks for testing. I agree, perfect it is not.

Considering all the discussed aspects, I feel this is the best possible implementation at this stage. I will go ahead and merge it, and start working on the user settings.

DannyBen:
@hojerst btw - when an inline indentation is needed, wouldn't the [originally proposed solution](https://github.com/DannyBen/bashly/tree/a654720873f5ecbebd5e5b305634bbc6078caba0/examples/heredoc#srclibheredocssh) work? The one where you write your function-that-returns-heredoc inside one of the `lib/*` files?

hojerst:
Maybe not indenting the code in the wrapper function would be the most compatible way (but this results in a wrongly indented generated script)

DannyBen:
Yeah, well if you want to wrap and use heredocs and indent your heredocs - the official answer is "touch luck" :)

I am curious - why use wrap at all? Isn't bashly providing you with sufficient tools to generate the full result without the need to wrap it in a function?

hojerst:
Wrapping isn't a functionality I personally need. I just want to provide this function to the users of my script. This allows to source the script without executing it directly. Isn't this the main function of the `-w` flag anyway? I can live with not having it wrapped for now though.

On the other hand moving 90% of my business logic to a function in lib and passing args from command to the lib function easily kills readability of code. (one of the major features of bashly imho).

Maybe for clarity my specific usecase:
I'm writing a small bash helper that allows me to manage a Unifi USG router. The router provides SSH but is rather limited regarding available tooling. I want to call a `ssh hostname command` with `command` being a multiline shell script. This script in itself needs to do regex replacements with sed. To make things more complicated, the regex needs to include variable parts AND needs to use `'` and `"`. Being regex the need to use `\` is a must too. The most readable way to generate this script for me is something like:

```
script="$(cat <<-EOF
script goes here
EOF
)"
```

Needing to deal with double and tripple escaping `\`, `'` and `"` is just aweful without heredocs in bash.

hojerst:
yeah I considered those, but there are multiple downsides to both approaches.

1. separate file: I need to insert dynamic values into the script. With heredocs you can just insert vars from the bash environment using ${var} syntax. Thats not trivial for external files. Sure I could copy them as is to the remote and run them with a argument, but this needs a temporary file on the server, which I do not want either.
2. for the usecase of needing vars expanded, I can't use single quote multiline strings (`'`). This leaves me with double quote (`"`) multiline strings. However as the regexes in my script contain a bunch of `\` and `"` by themself, I would need to escape those in the string. Something that is not needed for using heredocs.
As the generated script will also be passed to another host and run there, getting the quoting/escaping right is very important - and this is something already quite tedious when dealing with regexes (which my script also includes as it needs to use complex `sed` commands). 

Heredocs are just the simplest/most readable solution for my usecase. (I use heredocs quite often as a 'poor-mans template engine'). 

Anyway, I got a solution for my usecase which allows me to stick with bashly (which I really like btw).

DannyBen:
Alright. A version will be released tomorrow. Thanks for the feedback.

Issue #352, opened 2023-02-28
Provide a third state to `strict` which omits `set -e`


halostatue:
### Description

I usually write my shell scripts *without* `set -e`, because exits too quietly. Instead, I have a pair of standard functions (`fail-unless` and `halt`) which let me control flow much better and provide better messaging. I find `set -e` dangerous precisely because it turns harmless failures like the example below into silent halts.

```bash
[[ "${args["--force"]} ]] || return
```

I was debugging a script yesterday that was halting for unexplained reasons that became clear when I saw the generated script included `set -e`. I was able to resolve this with `set +e`, but when I use these, I prefer to use `set +e -o pipefail` (I *might* include `-u`, but I generally declare every variable that I use).

IMO, the states could be:

- `strict: true`: `set -euo pipefail`
- `strict: false`: `set -eo pipefail`
- `strict: disabled`: `set -o pipefail`
- `strict: manual`: omitted entirely; could be `none`, but `manual` declares that I, as the script writer, will take the responsibility for those settings myself.

I do think that `set -o pipefail` should be part of the written defaults all the time; there’s almost never a time when you want to hide individual pipe command failures, but I’m less bothered about that.

halostatue:
I disagree about the benefits of `set -e`; as I noted, I had silent early termination which was substantially harder to debug than if the script had continued.

The BashFAQ also indicates that `set -e` provides a tissue instead of a safety blanket: https://mywiki.wooledge.org/BashFAQ/105 because there’s a lot of corner cases. This is one of the reasons why I have explicitly *not* used `set -e` on all but the simplest of shell scripts but instead explicitly check for errors on lines where failure or success matters.

If `set -e` cannot be removed from Bashly safely, then I’ll probably have to make a prefix/suffix for all of my functions to disable/enable it for anything that is in my code, because I find `set -e` unusable and untrustworthy. (And yes, I use a *lot* of subshell calls, which is one of the places where `set -e` fails fairly hard.)

I agree with you about the confusing options; I do think that some other configuration option for this would be better.

DannyBen:
Even in the link you posted, the very last Conclusion paragraph states two opposite opinions. On says use it, the other don't.

The `set -e` option in bash is the closest thing to how all other languages operate - when there is an error, exit. Non zero return code in bash is error.

That said - bashly should not fight you if you feel otherwise.
I will execute all specs without `set -e` first, see if things pass. If they do, then the door may be open to allowing users to `set` whatever they want.

[...]

PR #353, opened 2023-03-01
Change settings.strict to allow any string

DannyBen:
Allow `settings.strict` to accept any custom string, in case the developer does not want `set -e` nor `set -euo pipefail`.

addresses #352


halostatue:
This looks like it would suit my needs perfectly.

PR #355, opened 2023-03-02
Implement Support for Alternate Dependencies

### LINKED ISSUE #354: Support for Alternate Dependencies
### Description

I want to be able to specify alternate dependencies so that this dependency comment can be true:

```yml
dependencies:
  jq: |-
    Ensure that you have installed jq or gojq. If gojq is available, it will
    be preferred.

    https://stedolan.github.io/jq/download/
```

This could be supported in either array or dict configurations by adding support for a dict value:

```yml
dependencies:
  - { commands: [jq, gojq], message: "Ensure you have installed either jq or gojq." }
```

or

```yml
dependencies:
  jq:
    commands: [jq, gojq]
    message: "Ensure that you have installed either jq or gojq."
```

This would be a minimal change to the generated code, because `builtin command` already accepts multiple arguments and returns 0 if any one of them is found.

halostatue:
Resolves #354

The change to `config_validator.rb` is my least favourite part of this pass. However, it seems cleaner to reimplement the nested portions of `assert_hash` with specific handling. I suspect that some of this could be simplified, but I also made the simplifying assumption that if you’re specifying *one* hash value implementation for dependencies, you should specify them all that way.

The changes to `run.gtx` are required to prevent shellcheck failure since `$dependencies` is unused otherwise. This could be changed to stay within `parse_requirements`, but it would require `declare -gA dependencies` above `local result` with the same shellcheck or some otherwise mostly-harmless access of the variable. The `-gA` is required so that the variable is visible outside of `parse_requirements`, since `declare` inside of a function is the same as `local`.

The otherwise mostly-harmless access could be something like:

```
> if ((${#dependencies[@]} != {{ dependencies.size }})); then
>   printf "Missing dependencies\n" >&2
> fi
```

I have not updated any documentation as I’m not sure that this is the right implementation that you were thinking of. I will note that it *does work* in my tests.

### DIFF
--- a/lib/bashly/config_validator.rb
+++ b/lib/bashly/config_validator.rb
@@ -50,7 +50,19 @@ module Bashly
       when Array
         assert_array key, value, of: :string
       when Hash
-        assert_hash key, value, of: :string
+        assert_hash key, value
+
+        if value.any? { |_k, v| v.is_a?(Hash) }
+          value.each do |k, v|
+            subkey = "#{key}.#{k}"
+
+            assert_hash subkey, v, keys: %i[commands message]
+            assert_array "#{subkey}.commands", v['commands'], of: :string
+            assert_string "#{subkey}.message", v['message']
+          end
+        else
+          assert_hash key, value, of: :string
+        end
       else
         assert [Array, Hash].include?(value.class),
           "#{key} must be an array or a hash"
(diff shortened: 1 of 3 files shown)


DannyBen:
Thanks for this attempt, but the implementation I was thinking about is a bit mode modelled (more code, but also higher maintainability and testability). 

I need to do this myself - I will push something soon.

(but leave this PR open for now).

PR #356, opened 2023-03-02
Update dependencies option to support 'any' (e.g. wget or curl)

DannyBen:
cc: #355, #354

### DIFF
--- a/lib/bashly/concerns/validation_helpers.rb
+++ b/lib/bashly/concerns/validation_helpers.rb
@@ -35,18 +35,12 @@ module Bashly
       end
     end
 
-    def assert_hash(key, value, keys: nil, of: nil)
+    def assert_hash(key, value, keys: nil)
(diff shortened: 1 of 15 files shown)


DannyBen:
@halostatue - I believe the implementation is complete.

The behavior is nicely visible here:
https://github.com/DannyBen/bashly/pull/356/files#diff-118bd7cd9676740ea3db672f08af1c32623fafb878aab7509609d9ddaadda65f

halostatue:
This looks **spectacular**. The only (minor) comment that I have is that the documentation implies that you only get `$deps` when you use the alt hash format, but `$deps` is always present. I’m not sure if it’s worth documenting `$deps` beyond what is there, but I will be using it.

DannyBen:
Thanks. Fixed.

I will merge this, and update the bashly site documentation to reflect, and after that a release?

Issue #368, opened 2023-04-07
`allowed` incompatible with `required: false`


ChipS3t:
### Bashly Version

Latest Ruby Gem

### Description

When using the "allowed" option for args or flags, the arg or flag becomes required, throwing an error if a value isn't supplied, even with "required: false" set.

The docs explicitly [states](https://bashly.dannyb.co/configuration/argument/#allowed) the `allowed` option should work with the `required` option, but it does not. Setting a default value is not always preferable when using `allowed`. For instance, imagine a scenario where a command has a set of default actions it takes. If you wanted the user to be able to exclude one or more of those actions, you could have an `--exclude` flag followed by a list of allowed actions to exclude from the script's execution. However, by default, you wouldn't want to exclude any options; therefore, it doesn't make sense to set a default value for the flag. You could add a "none" option, but this is clunky and doesn't really make sense as an allowed option (if I wanted to not exclude anything, I shouldn't have to pass the option at all). If the user attempted to pass an invalid value to the flag or argument, "none" would be printed alongside the valid options.

This works as expected if `repeatable: true` option is used.

### Contents of bashly.yml

```yaml
name: download
help: Sample minimal application without commands
version: 0.1.0

args:
- name: source
  help: URL to download from
  required: false
  allowed: [sphinx, abc, wonder]

flags:
- long: --force
  short: -f
  arg: force
  help: Overwrite existing files
  required: false
  allowed: [a, b, c]
```

### Reproduction Steps

`bashly init --minimal`
Edit `bashly.yml` to look like the provided example.
`bashly generate`
`./download`

### Actual Behavior

Outputs: ` source must be one of: sphinx, abc, wonder` or `--force must be one of: a, b, c`.

### Expected Behavior

Successful execution with no output.

DannyBen:
Well, this is by design.

Setting the `allowed` whitelist, means the variable you receive in your script will always only be one of the allowed options.
The docs are not wrong in stating it can be used with either `default` or `required`. There is no need to ever specify `required: false`, only `required: true` since false is the default.

In your use case, allowing the `--exclude` flag to receive a `none` value, is the way to solve this.
The use does not need to specify `--exclude none`, since you make it the default.

```yaml
name: download
help: Sample minimal application without commands
version: 0.1.0

args:
- name: source
  help: URL to download from
  default: sphinx
  allowed: [sphinx, abc, wonder]

flags:
- long: --exclude
  short: -e
  arg: action
  help: Avoid executing this action
[...]

PR #372, opened 2023-04-07
Fix `arg.allowed` and `flag.allowed` to allow setting without `required` or `default`

DannyBen:
cc #368

This is a revert of #370 plus a fix to allow the whitelist to be optional and without a default value.


lib/bashly/views/command/whitelist_filter.gtx
DannyBen, line 1: Modify the whitelist filter to only validate the contents of the variable if it is not empty.

In case the flag/arg are set to `require: true`, we will never reach this code since it will be blocked earlier.

PR #378, opened 2023-04-17
Feature/add json schema

EmilySeville7cfg:
image

Later you can validate YAML files via some schema validator in Ruby to simplify validation code.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,36 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "title": "cli",
+    "description": "A CLI of the current application",
+    "type": "object",
+    "properties": {
+        "name": {
+            "title": "name",
+            "description": "A name of the current application",
+            "type": "string",
+            "minLength": 1,
+            "examples": [
+                "download"
+            ]
+        },
+        "help": {
+            "title": "help",
+            "description": "A description of the current application",
+            "type": "string",
+            "minLength": 1,
+            "examples": [
+                "Sample minimal application without commands"
+            ]
+        },
+        "version": {
+            "title": "version",
+            "description": "A version of the current application",
+            "type": "string",
+            "minLength": 1,
+            "examples": [
+                "0.1.0"
+            ]
+        }
+    },
+    "additionalProperties": false
+}
--- a/schema.json
+++ b/schema.json
@@ -1,5 +1,453 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema",
+    "definitions": {
+        "command": {
+            "title": "command",
+            "description": "A command of the current application",
+            "type": "object",
+            "required": [
+                "name"
+            ],
+            "properties": {
+                "name": {
+                    "title": "name",
+                    "description": "A name of the current command",
+                    "type": "string",
+                    "minLength": 1,
+                    "examples": [
+                        "source"
+                    ]
+                },
+                "alias": {
+                    "title": "alias",
+                    "description": "An alias of the current command",
+                    "oneOf": [
+                        {
+                            "type": "string",
+                            "minLength": 1,
+                            "examples": [
+                                "s"
+                            ]
+                        },
+                        {
+                            "type": "array",
+                            "minItems": 1,
+                            "uniqueItems": true,
+                            "items": {
+                                "type": "string",
+                                "minLength": 1,
+                                "examples": [
+                                    "s"
+                                ]
+                            }
+                        }
+                    ]
+                },
+                "help": {
+                    "title": "help",
+                    "description": "A description of the current command",
+                    "type": "string",
+                    "minLength": 1,
+                    "examples": [
+                        "URL to download from"
+                    ]
+                },
+                "args": {
+                    "title": "args",
+                    "description": "Arguments of the current command",
+                    "type": "array",
+                    "uniqueItems": true,
+                    "minItems": 1,
+                    "items": {
+                        "$ref": "#/definitions/argument"
+                    }
+                },
+                "flags": {
+                    "title": "flags",
+                    "description": "Flags of the current command",
+                    "type": "array",
+                    "uniqueItems": true,
+                    "minItems": 1,
+                    "items": {
+                        "$ref": "#/definitions/flag"
+                    }
+                },
+                "commands": {
+                    "title": "commands",
+                    "description": "Subcommands of the current command",
+                    "type": "array",
+                    "uniqueItems": true,
+                    "items": {
+                        "$ref": "#/definitions/command"
+                    }
+                },
+                "default": {
+                    "title": "default",
+                    "description": "Whether to use the current command as a fallback when CLI used incorrectly",
+                    "type": "boolean",
+                    "default": false
+                },
+                "environment_variables": {
+                    "$ref": "#/definitions/environment_variables"
+                },
+                "examples": {
+                    "title": "examples",
+                    "description": "An example of the current command",
+                    "oneOf": [
+                        {
+                            "type": "string",
+                            "minLength": 1,
+                            "examples": [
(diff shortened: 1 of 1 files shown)


DannyBen:
Hmm. That's nice.
There are a couple reasons I opted for home-made validation:

1. JSON schema seems too verbose.
2. Bashly's validation has some special cases (that may or may not be solvable with JSON schema, I was lazy to check).
3. I wanted to fully control error messages.

Some questions:

1. Does this schema currently cover all the options in bashly.yml?
2. Does it mimic the same object structure as Bashly's classes (Command / Argument / Flag etc.)?
3. To your knowledge, will we be able to replicate all the validations that exist in [`Bashly::ConfigValidator`](https://github.com/DannyBen/bashly/blob/master/lib/bashly/config_validator.rb)?

EmilySeville7cfg:
> Does this schema currently cover all the options in bashly.yml?

I hope it will today, or tomorrow. There are some questionable things, like `environment_variables` inside top-level `commands`, I have to check whether everything is correct before tell that it covers everything. :)

> Does it mimic the same object structure as Bashly's classes (Command / Argument / Flag etc.)?

What do you mean by it? I just follow documentation and examples and try to make schema to pass for all of them and fail for some wrong cases.

> To your knowledge, will we be able to replicate all the validations that exist in [Bashly::ConfigValidator](https://github.com/DannyBen/bashly/blob/master/lib/bashly/config_validator.rb)?

From my understanding, it can't, but most of the things can be covered by JSON schema validator. To be clear, there are some issues with JSON schema validators I use. They are not fatal, but may cause some inconvenience. For instance, VS Code extension validator doesn't check whether two objects are equal in an array, just plain values are checked for equality.

DannyBen:
> Does it mimic the same object structure as Bashly's classes (Command / Argument / Flag etc.)?

What I mean by that is this (and it is also related to your environment_variables in the root comment):

a) The entire bashly.yml represents a bashly `Command` object. As can be seen in the validation "entry point" command:
https://github.com/DannyBen/bashly/blob/7b90f7275e42477453962e94718dade3dc1343bc/lib/bashly/config_validator.rb#L11-L13

b) The command object can include arrays of other objects, such as `Flag`, `Argument`, `Dependency`, `EnvironmentVariable`, `Example` etc. Can be seen here:
https://github.com/DannyBen/bashly/blob/7b90f7275e42477453962e94718dade3dc1343bc/lib/bashly/config_validator.rb#L182-L187

So, in other words, when modelling the schema for an array of flags (for example), we do not say "array of hashes that include these keys", we say "array of flags" instead, and "flag" is a custom type that has its own validations.

---

The fact that this validation cannot cover all the cases that the validator currently covers, makes it a possible show stopper. It means that we will either need to compromise, and avoid failing on things that the schema validator cannot process, or have two different validations - one which is the currently implemented validator, and another, which is the schema validator intended for VS code users or other IDEs that implement inline validation.

Thoughts?

EmilySeville7cfg:
> The fact that this validation cannot cover all the cases that the validator currently covers, makes it a possible show stopper. It means that we will either need to compromise, and avoid failing on things that the schema validator cannot process, or have two different validations - one which is the currently implemented validator, and another, which is the schema validator intended for VS code users or other IDEs that implement inline validation.

I suggest to use JSON schema validator and add some custom validations on top of it where validator can detect errors (like with object equality). Maybe there is a validator which allows to customize error messages.

And I don't think that it's a huge problem that VS Code validation is more permissive than built-in your tool directly. For example, there is JSON schema for GitHub workflow file in SchemaStore, but it can't check whether variable expansion `${{ ... }}` syntax is right inside YAML, it's too complex check for JSON schema validator to make it possible to implement. In other words schema by itself allows incorrect variable expansions, but GitHub Actions what read YAML files don't and it's fine. JSON schema is not designed to do such complex stuff for now.

DannyBen:
For the CI, I found [`check-jsonschema`](https://github.com/python-jsonschema/check-jsonschema) to be easy to use. Putting it here for reference.

```
$ pip install check-jsonschema
$ check-jsonschema --schemafile bashly-schema.json bashly.yml
```

EmilySeville7cfg:
@DannyBen, bashly.yml JSON schema is ready to test. :) There is just one thing left to do - add links to docs for all properties, I guess I can do it tomorrow.

DannyBen:
Thanks, I will take a look. A few random thoughts:

1. Since this is not going to influence the CLI validation, and since there are already expansion thoughts - I wonder if this schema should live in this repository, or in its own.
2. Personally, I have never needed to have a JSON schema, which means I will probably not be good at maintaining it - this adds another point related to the previous one.
3. Since manually editing JSON is much more error prone than editing YAML - I wonder if we should use YAML as the source foe the schema. 
   - The `check-jsonschema` CLI supports YAML as input
   - We can add a simple utility command for developers, using the [Runfile](https://github.com/DannyBen/bashly/blob/master/runfile) to convert the YAML schema to JSON.

DannyBen:
> I can maintain it, I often write JSON schemas. 

Cool. This is good to know.

> I am not sure whether VS Code editor validator supports JSON schemas passed in YAML format

It wont need to, since we will always have both the YAML and the JSON in the repo. But if you are maintaining it, then it is your call. I can easily create a `run schema-convert` command which will read the YAML and write the JSON.

---

These two examples currently fail:

- examples/dependencies-alt/src/bashly.yml
- examples/split-config/src/bashly.yml

DannyBen:
I added `check-jsonschema` to my [rush repo](https://github.com/DannyBen/rush-repo/tree/master/check-jsonschema), so that we can later just install it in the workflow and run `run schema` to run the tests.

https://github.com/DannyBen/bashly/blob/7b90f7275e42477453962e94718dade3dc1343bc/.github/workflows/test.yml#L50

DannyBen:
@EmilySeville7cfg - I suggest not doing the settings schema in this PR.
I would like it to be merged as soon as you are ready with just the bashly.yml, then I can add the workflow, and we lock one cycle before moving to the next.

Will it be more convenient if I add you as collaborator?

EmilySeville7cfg:
> @EmilySeville7cfg - I suggest not doing the settings schema in this PR.
I would like it to be merged as soon as you are ready with just the bashly.yml, then I can add the workflow, and we lock one cycle before moving to the next.

Not an issue. :) It's more convenient for me too.

> Will it be more convenient if I add you as collaborator?

Definitely will. :)

schema.json
```diff
@@ -0,0 +1,810 @@
...
+                    "default": true
+                },
+                "allowed": {
+                    "title": "allowed",
+                    "description": "Valid values of the current flag",
+                    "type": "array",
+                    "minLength": 1,
+                    "uniqueItems": true,
+                    "items": {
+                        "description": "A valid values of the current positional argument",
```
DannyBen, line 153: "A valid values"

schema.json
```diff
@@ -0,0 +1,810 @@
...
+                    }
+                },
+                "completions": {
+                    "title": "completions",
+                    "description": "Completions of the current flag",
+                    "type": "array",
+                    "minLength": 1,
+                    "uniqueItems": true,
+                    "items": {
+                        "description": "A completions of the current flag",
```
DannyBen, line 183: "A completions"

DannyBen (changes requested):

schema.json
EmilySeville7cfg: Fixed

DannyBen:
> what about to move 2nd and 3rd task

Sure thing. Let me know when you are done just with the schema, I will then merge and add the CI myself.

DannyBen (approved):

EmilySeville7cfg:
I think schema is ready for production. Further enhancements can be done, but not today. 😄

PR #380, opened 2023-04-18
Feature: schema for settings.yml

EmilySeville7cfg:
I suggest permitting use **bashly-settings.yml** name also or something like that to clearly distinguish this config from numerous others. **settings** is a too generic name and for this reason it can't be added to SchemaStore. It means that users should manually enable validation for **settings.yml** in their editor config like **.vscode/settings.json**:

```yaml
{
    "yaml.schemas": {
        "path/to/this-schema.json": "path/to/settings.yml"
    }
}
```


schemas/settings.json
```diff
@@ -0,0 +1,144 @@
...
+                "blue_bold",
+                "magenta_bold",
+                "cyan_bold",
+                "red_underlined",
+                "green_underlined",
+                "yellow_underlined",
+                "blue_underlined",
+                "magenta_underlined",
+                "cyan_underlined",
+                "~"
```
DannyBen, line 27: An unquoted `~` means null in YAML - is it ok to define it like a string in the schema?

schemas/settings.json
```diff
@@ -0,0 +1,144 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "definitions": {
+        "color": {
+            "type": "string",
+            "enum": [
```
DannyBen, line 6: I am not even sure that this setting should have a whitelist.
The user can create any color function they want, and then place its name here.

schemas/settings.json
EmilySeville7cfg: Otherwise, this value will not be accepted while docs [list](https://bashly.dannyb.co/usage/settings/#usage_colors) it as the default valid one. So yeah, it's okay. From this perspective. Another question how extension being used inserts such string when suggesting it - in quotes or not. VS Code for instance put it unquoted what results in type error telling that string expected, but it's an extension issue. :)

schemas/settings.json
EmilySeville7cfg: What color formats are permitted? RBG maybe too?

DannyBen:
> I suggest permitting use bashly-settings.yml 

Not sure about this. The user's working directory is only intended for creating a bashly project. Having to repeat the name in the settings is redundant from any other standpoint.

schemas/settings.json
DannyBen: No. Any string is allowed here, and the user is expected to have a function with the same name.
The colors you outlined in the list, are just the colors that the [built in color library](https://github.com/DannyBen/bashly/blob/master/lib/bashly/libraries/colors/colors.sh) brings.

Users don't have to use it, they can have a function named `command_color` and this will be the string that they will use in the colors in settings.

EmilySeville7cfg: Now any color is permitted, but standard ones are suggested.

EmilySeville7cfg:
> Not sure about this. The user's working directory is only intended for creating a bashly project. Having to repeat the name in the filename is redundant from any other standpoint.

Then, maybe use just **strings.yml** instead of **bashly-strings.yml**?

schemas/settings.json
DannyBen: Perfect

DannyBen:
> Then, maybe use just strings.yml instead of bashly-strings.yml?

Hmm..... you got me there. Lets re-evaluate and discuss in a separate issue?

EmilySeville7cfg:
> Lets re-evaluate and discuss in a separate issue?

#381

DannyBen:
Testing the default settings.yml (`bashly add settings`) with the schema fails:

```
$ check-jsonschema --schemafile settings.json settings.yml
Schema validation errors were encountered.
  settings.yml::$.usage_colors.caption: None is not of type 'string'
  settings.yml::$.usage_colors.command: None is not of type 'string'
  settings.yml::$.usage_colors.arg: None is not of type 'string'
  settings.yml::$.usage_colors.flag: None is not of type 'string'
  settings.yml::$.usage_colors.environment_variable: None is not of type 'string'
```

This seems to be the same issue we tackled earlier with the null. `None` is null in python.
I think we should not use the VSCode extension as a guide - if it has bugs, they need to fix it.

DannyBen:
I added the settings schema test, which fails as expected.
Sorry for stepping on your toes :)

EmilySeville7cfg:
> I added the settings schema test, which fails as expected.

Maybe it should pass then when it fails? :)

DannyBen:
> Maybe it should pass then when it fails? :)

It should allow `null`

DannyBen:
Ready to merge?
I want to perhaps separate the static analysis tests from the JSON schema tests (post merge).

EmilySeville7cfg:
Ready :)

Issue #398, opened 2023-07-03
Organize scripts in src dir in subdirectories


n-rodriguez:
### Description

Hi there!

Today scripts are organized in `src`dir in a flat way :

```sh
src/
├── bashly.yml
├── completions_command.sh
├── concerto_config_dump_command.sh
├── concerto_config_print_command.sh
├── concerto_notify_release_command.sh
├── concerto_notify_sentry_command.sh
├── concerto_restart_command.sh
├── concerto_setup_command.sh
├── concerto_update_french_logo_command.sh
├── concerto_update_staging_command.sh
├── docker_config_command.sh
├── docker_down_command.sh
├── docker_enter_command.sh
├── docker_exec_command.sh
├── docker_images_command.sh
├── docker_logs_command.sh
├── docker_prune_images_command.sh
├── docker_ps_command.sh
├── docker_pull_command.sh
├── docker_restart_command.sh
├── docker_start_command.sh
├── docker_stop_command.sh
├── docker_top_command.sh
├── docker_up_command.sh
├── get_config_concerto_version_command.sh
├── get_config_date_command.sh
├── get_config_debian_version_command.sh
├── get_config_env_command.sh
├── get_config_local_command.sh
├── get_config_service_version_command.sh
├── lib
│   ├── colors.sh
```

It would be great to have an option to organize scripts in subdirectories :

```sh
src/
├── bashly.yml
├── completions_command.sh
├── concerto
│   └── config_dump_command.sh
│   └── config_print_command.sh
│   └── notify_release_command.sh
│   └── notify_sentry_command.sh
│   └── restart_command.sh
│   └── setup_command.sh
│   └── update_french_logo_command.sh
│   └── update_staging_command.sh
├── docker
│   └── start_command.sh
│   └── config_command.sh
│   └── down_command.sh
│   └── enter_command.sh
│   └── exec_command.sh
│   └── images_command.sh
│   └── logs_command.sh
│   └── prune_images_command.sh
│   └── ps_command.sh
│   └── pull_command.sh
│   └── restart_command.sh
│   └── start_command.sh
│   └── stop_command.sh
│   └── top_command.sh
│   └── up_command.sh
```

Thank you!

DannyBen:
Well - it is nice to see that bashly is easy enough to use that you have built (or planning to build) such a large bash script with it.

There is a `command.filename` option which lets you specify the path (in `src`) for each command. Isn't this sufficient?

See:

[...]

PR #399, opened 2023-07-03
Add suport for auto-organizing command files in subfolders

DannyBen:
cc #398


DannyBen:
@EmilySeville7cfg - this PR contains a schema change to the settings.yml file:
👉 https://github.com/DannyBen/bashly/pull/399/commits/40ddcb1cdf5d7ce42d704085251163823563bc82


Can you verify my change?
The new Settings option is `commands_dir`, which is null (`~`) by default, and accepts any string.

schemas/settings.json
```diff
@@ -66,9 +66,22 @@
             "minLength": 1,
             "default": "lib"
         },
+        "commands_dir": {
+            "title": "commands dir",
+            "description": "A directory to nested command files of the current script",
+            "oneOf": [
```
EmilySeville7cfg, line 72: I guess it's better to put schemas containing `"type": "null"` at the end, not to introduce inconsistency. But overall this change looks nice. ;)

EmilySeville7cfg (approved):

DannyBen:
Change as you see fit - schemas are your domain :)

EmilySeville7cfg:
https://bashly.dannyb.co/usage/settings/#commands_dir is not a valid link. 🤔

DannyBen:
> https://bashly.dannyb.co/usage/settings/#commands_dir is not a valid link. 🤔

Use it. It will be live when the new version is released.

PR #400, opened 2023-07-03
Add doc references and add examples for <> values

EmilySeville7cfg:
- [x] bashly
- [x] settings
- [x] strings
- [ ] examples for [**completions**](https://bashly.dannyb.co/advanced/bash-completion/#custom-command-completions) property

## Notes

Regex used to speed up adding doc references:

- from: `"title": "(.*)",\n(\s+)"description": "(.*)",`
- to: `"title": "$1",\n$2"description": "$3\\nURL/#$1",`


DannyBen:
Do you prefer that I merge #399 so you have my latest schema?

EmilySeville7cfg:
> Do you prefer that I merge https://github.com/DannyBen/bashly/pull/399 so you have my latest schema?

@DannyBen, yes, please.

EmilySeville7cfg:
@DannyBen, everything is ready.

DannyBen:
> is there any list of all possible special values for [**completions**](https://bashly.dannyb.co/advanced/bash-completion/#custom-command-completions) property like `<user>`?

Yes. 
https://github.com/DannyBen/completely#suggesting-files-directories-and-other-bash-built-ins

But - notice that there may be other possibilities that are not listed, since these types of strings are simply used differently. From the completely page (appears below that list):

> For those interested in the technical details, any word between <...> will simply be added using the [compgen -A action](https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html) function, so you can in fact use any of its supported arguments.

So - if there is a way to suggest this list, but still allow any other `<...>` option - it is okay.

Let me know if wait or merge.

EmilySeville7cfg:
You can merge.

DannyBen (approved):

DannyBen:
Nice. Thanks for this.

I wonder why JSON schema doesn't have a dedicated field for reference/documentation. Seems basic.

EmilySeville7cfg:
> I wonder why JSON schema doesn't have a dedicated field for reference/documentation. Seems basic.

The same question. I think they can introduce something like `documentation` property for this purpose, but it will be useless for VS Code users until any extension supports it.

PR #404, opened 2023-07-22
Place `config.ini` in the executable dir

dvrkn:
Hi! Thank you for the great tool!

We are using it putting the executable to the $PATH, and expecting to config.ini will be always placed the same dir as the executable.

Here is the little fix. Its my first one, so sorry for inconvenience if any :)

### DIFF
--- a/lib/bashly/libraries/config/config.sh
+++ b/lib/bashly/libraries/config/config.sh
@@ -13,7 +13,8 @@
 ##
 config_init() {
   CONFIG_FILE=${CONFIG_FILE:=config.ini}
-  [[ -f "$( dirname -- "$0"; )/$CONFIG_FILE" ]] || touch "$( dirname -- "$0"; )/$CONFIG_FILE"
+  EXECUTABLE_PATH=$( dirname -- "$0"; )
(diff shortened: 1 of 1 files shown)


DannyBen:
Thanks for this PR, but this change is not necessary, here is why:

### A) Library functions are a "starting point"

Everything added by `bashly add *` is provided as a starting point, with the intention of allowing the developer to edit them as they see fit. Just if you edit any file, be sure to remove the "update magic comment"  from the first line (`[@bashly-upgrade config]`) - to avoid accidentally overriding this file when running `bashly generate --upgrade`

### B) Use `initialize.sh`

The proper way to set the config file location for your script is to use the `src/initialize.sh` file. You can run `bashly add hooks` to get three hook files (before / after / initialize) - and delete whatever you don't need.

In this file, simply place this (or whatever you want).

```bash
CONFIG_FILE="$(dirname -- "$0")/config.ini"
```

### C) Recommended config location

I would like to recommend against placing the config file alongside the executable script. This is rarely (never?) done by any other program. Config files normally go in the user's `~/.config/` directory, or in a global system config directory, or in `~/.appname/config.ini`.

dvrkn:
thank you!

DannyBen:
You're welcome.

Issue #426, opened 2023-08-26
Allow having arbitrary keys (`x-anything`) in `bashly.yml`


DannyBen:
### Description

This can be useful for custom renderers, for example to add a `x-manpage-description` string for manpage renderes, and so on.

This should also be included in the JSON schema (cc @EmilySeville7cfg).

I think this should help, but can't seem to place it properly in the schema for it to stop bugging me:

```json
"patternProperties": { "^x-": {} }
```

This bashly.yml should pass schema check:

```yaml
name: docker
x-string: hello
x-hash: { key: value }
x-array: [1, 2]
x-int: 1
x-float: 1.2
x-boolean: yes
x-nil: ~

environment_variables:
- name: rush_config
  x-any: hello

commands:
- name: status
  x-any: hello

  args:
  - name: repo
    x-any: hello

  flags:
  - long: --purge
    x-any: hello

- name: images
  commands:
  - name: ls
    x-any: hello
```

PR #427, opened 2023-08-26
Allow arbitrary (`x-anything`) keys in bashly.yml

DannyBen:
cc #426

- [x] The `ConfigValidator` now ignores hash keys that start with `x-`
- [x] The JSON schema now allows `x-` keys in all the primary objects.

@EmilySeville7cfg - can you verify this is the best way to allow `x-anything`? I only added it in several places, assuming there is no way to allow it *everywhere*.

### DIFF
--- a/schemas/bashly.json
+++ b/schemas/bashly.json
@@ -1,6 +1,17 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema",
     "definitions": {
+        "custom-properties": {
+            "patternProperties": {
+                "^x-.": {
+                    "title": "custom property",
+                    "description": "A custom property",
+                    "examples": [
+                        "hello"
+                    ]
+                }
+            }
+        },
         "argument": {
             "title": "argument",
             "description": "A positional argument of the current script or sub-command\nhttps://bashly.dannyb.co/configuration/argument/",
@@ -74,8 +85,8 @@
                     ]
                 }
             },
-            "additionalProperties": false,
-            "patternProperties": { "^x-.": {} }
+            "$ref": "#/definitions/custom-properties",
+            "additionalProperties": false
         },
         "flag": {
             "title": "flag",
@@ -210,8 +221,8 @@
                     ]
                 }
             },
-            "additionalProperties": false,
-            "patternProperties": { "^x-.": {} }
+            "$ref": "#/definitions/custom-properties",
+            "additionalProperties": false
         },
         "name-property": {
             "title": "name",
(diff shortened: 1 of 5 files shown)


schemas/bashly.json
```diff
@@ -74,7 +74,8 @@
                     ]
                 }
             },
-            "additionalProperties": false
+            "additionalProperties": false,
+            "patternProperties": { "^x-": {} }
```
EmilySeville7cfg, line 78:
```suggestion
            "patternProperties": { "^x-.": {} }
```

It's better to use `.` not to allow properties consisting of just prefix `x-`.

EmilySeville7cfg (changes requested):

schemas/bashly.json
DannyBen: I don't understand.

DannyBen: Sure. I have updated the regex, feel free to push to this branch with your updates.

Example would be providing another "help" for a command, flag or arg, that will be used for man page generation:

```yaml
commands:
- name: status
  help: Show status
  x-manpage-help: |-
    This command shows the status of all `images`, `containers` and `volumes`.
    Use with `--verbose` to see deleted containers.
```

EmilySeville7cfg:
Everything is ready. :)

schemas/bashly.json
```diff
@@ -1,6 +1,17 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema",
     "definitions": {
+        "custom-properties": {
+            "patternProperties": {
+                "^x-.": {
+                    "title": "custom property",
+                    "description": "A custom property",
```
DannyBen, line 8: 1. Can we say "A custom property or object" to hint that we support anything?
2. Does the example "hello" really add anything? I would prefer without it.

schemas/bashly.json
```diff
@@ -74,6 +85,7 @@
                     ]
                 }
             },
+            "$ref": "#/definitions/custom-properties",
```
DannyBen, line 88: Much better as a `$ref`, thanks. I wasn't sure if I can do that.

schemas/bashly.json
EmilySeville7cfg: 1. I guess it's better to say `A custom property of any type`.
2. I've updated the example to `This command shows the status of all `images`, `containers` and volumes. Use with --verbose to see deleted containers.`.

DannyBen:
Can you approve?
I will probably change the example. If you feel it must be there, then maybe the word "anything" or "..." - anything else is confusing.

EmilySeville7cfg (approved):

DannyBen:
Excellent - thanks for the help.

I will change the example and probably change it to be `x_` and not `x-` - it is easier then for users to request the property by calling `command.x_anything`.

EmilySeville7cfg:
Yep, I agree, underscore maybe a better alternative, to keep all property names consistent - without dashes.

PR #431, opened 2023-08-27
Add config generator

EmilyGraceSeville7cf:
This simple script allows to interactively create configs with specified flags. While creating config via it user can add/remove flags, preview config before writing it into file.

Required dependencies:

- https://github.com/charmbracelet/gum: for interactive interaction
- https://github.com/kigster/dupervisor: for JSON -> YAML conversion
- https://github.com/sharkdp/bat: for config preview


DannyBen:
This is very meta... a bash script that creates a YAML config that creates a bash script... :)

This is not the bashly way.

EmilySeville7cfg:
I guess then you can close this PR.

DannyBen:
Yes, sorry. The `bashly` command is the only user-facing entrypoint and editing YAML files (or starting with an example) is easy.

PR #436, opened 2023-09-27
Integrate with unofficial draft-07 schema for better hints and validation

EmilyGraceSeville7cf:
Why to make this change is explained [here](https://github.com/SchemaStore/schemastore/pull/3260). Note that this schema may produce false positives like when `$ref` is used and `title`/`description`/`type` are not stated directly, but those cases can be safely ignored. It's better to get more errors when most of them are right, than get less errors and miss some real issues. :) This schema is designed to work the best with VS Code as tested just through YAML Red Hat extension.

image


DannyBen (approved):
I typically avoid "unofficial" anything, but since I am not a consumer of this JSON schema, I will defer to your judgment.

PR #464, opened 2023-12-19
Support array in default

### LINKED ISSUE #463: Add support for array in `flag.default` and `arg.default`
@EmilyGraceSeville7cf - got another one for you, if you can.

https://github.com/DannyBen/bashly/pull/462 added support for having an array in the `default` option of both `arg` and `flag`.

The array form is only allowed when `repeatable` is true (but I leave it up to you to decide if you wish to incorporate this limitation, or just allow both strings and arrays).

Funnily enough, the schema validation test [passes as it is now](https://github.com/DannyBen/bashly/actions/runs/7263119625/job/19787719856), for reasons I do not understand.

The following bashly.yml locations **should fail now** (since they are arrays and not strings), and pass after you do your thing:

https://github.com/DannyBen/bashly/blob/f45b99368e682d774ca38e2054d51033d168414c/examples/repeatable-arg/src/bashly.yml#L20-L22

https://github.com/DannyBen/bashly/blob/f45b99368e682d774ca38e2054d51033d168414c/examples/repeatable-flag/src/bashly.yml#L28-L30

The tests are done with the latest `check-jsonschema` CLI version 0.27.3

EmilyGraceSeville7cf:
closes #463

The easiest way to add array support for `default`.

To implement dependency between `repeatable` and `default` properties here `if` is required, but it means that common properties to `then` and `else` branches should be declared inside `definitions`. Will u accept such change if I make it?

### DIFF
--- a/examples/catch-all/src/bashly.yml
+++ b/examples/catch-all/src/bashly.yml
@@ -8,8 +8,7 @@ catch_all: true
 args:
 - name: message
   required: true
-  default: [""]
-  x_a: 
+  help: Message
 
 flags:
 - long: --debug
--- a/examples/catch-all/src/bashly.yml
+++ b/examples/catch-all/src/bashly.yml
@@ -8,7 +8,8 @@ catch_all: true
 args:
 - name: message
   required: true
-  help: Message
+  default: [""]
+  x_a: 
 
 flags:
 - long: --debug
--- a/schemas/bashly.json
+++ b/schemas/bashly.json
@@ -41,7 +41,18 @@
                 "default": {
                     "title": "default",
                     "description": "A default value of the current positional argument\nhttps://bashly.dannyb.co/configuration/argument/#default",
-                    "type": "string",
+                    "oneOf": [
+                        {
+                            "type": "string"
+                        },
+                        {
+                            "type": "array",
+                            "minItems": 1,
+                            "items": {
+                                "type": "string"
+                            }
+                        }
+                    ],
                     "examples": [
                         "*.jpg"
                     ]
@@ -159,7 +170,18 @@
                 "default": {
                     "title": "default",
                     "description": "A default value of the current flag\nhttps://bashly.dannyb.co/configuration/flag/#default",
-                    "type": "string",
+                    "oneOf": [
+                        {
+                            "type": "string"
+                        },
+                        {
+                            "type": "array",
+                            "minItems": 1,
+                            "items": {
+                                "type": "string"
+                            }
+                        }
+                    ],
                     "examples": [
                         "*.jpg"
                     ]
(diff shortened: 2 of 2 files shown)


DannyBen:
I think what you have done here is sufficient, no?

Have you figured out how come the `check-jsonschema` did not fail when only string is allowed, and it was an array?

EmilyGraceSeville7cf:
> I think what you have done here is sufficient, no? I prefer simplicity whenever possible.

Got it. ;)

Have you figured out how come the check-jsonschema did not fail when only string is allowed, and it was an array?

It may be a bug, I'll check `check-jsonschema` and some other validators. I see that `rush` command is used to install `check-jsonschema`. I guess it is [this](https://rushjs.io/pages/intro/welcome/) one, am I right?

DannyBen:
> I guess it is [this](https://rushjs.io/pages/intro/welcome/) one, am I right?

Nope. It is [my own rush](https://github.com/DannyBen/rush-cli), and the installation script for check-jsonschema is in my own [rush repo](https://github.com/DannyBen/rush-repo/tree/master/check-jsonschema).

[I simply install it with pipx](https://github.com/DannyBen/rush-repo/blob/f21adb43cfe5f7e447ee02f7b72e622bfed181f2/check-jsonschema/main-arch#L11) (on my arch linux).


```
$ pipx install check-jsonschema
```

(sorry - just realized I edited your comment instead of replying so I recreated yours.... 🤦)

EmilyGraceSeville7cf:
It seems to be a bug, I have to investigate why it happens and report about it if developers aren't already informed about it.

DannyBen:
So - is this ready to merge from your standpoint?

EmilyGraceSeville7cf:
> So - is this ready to merge from your standpoint?

Yep.

DannyBen:
... and one last thing:

> It seems to be a bug, I have to investigate why it happens and report

If you know of a better, or the "de facto standard" CLI for validating JSON schema, let me know and I will switch to it.

DannyBen:
> I suggest create an issue about this CI problem with check-jsonschema not to forget about it until it fixed.

Right. I will, once I am also able to create a reproducible simple test case.

DannyBen:
Thanks for your quick implementation, as always.

DannyBen (approved):

DannyBen:
@EmilyGraceSeville7cf - see this:

The following schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "default": {
      "type": "string"
    }
  },
  "required": ["default"],
  "additionalProperties": false
}
```

with this test file:

```yaml
# default: some string
default:
  - one
  - two
```

**fails as expected** when running this command:

```
$ check-jsonschema --schemafile schema.json test.yml
```

which makes me wonder, if it is a problem in our schema and not in the validator itself.

Issue #466, opened 2023-12-19
Add JSON schema tests to ensure the validation fails when it should


DannyBen:
### Description

Based on the findings in #465, we should have our JSON schema CI tests also run against a few invalid YAML configurations to ensure the schema fails when it should.

DannyBen:
Done.

PR #467, opened 2023-12-19
Improve JSON schema tests

DannyBen:
Blocked by #466 

This PR adds the necessary infrastructure for easily ensuring that invalid JSON schemas fail as expected.
For the time being, there is one single YAML file for each of the schemas (settings, strings, bashly).

As expected, the bashly schema fails. When it passes, we should probably add at least a handful more test files.

---

- [x] Add a `spec/fixtures/schema_invalid` folder, with subfolder-name-as-schema convention, and with test YAMLs in each.
- [x] Add `run schema invalid` test command that tests all invalid fixtures in `spec/fixtures/schemas_invalid`
- [x] Update `run schema all` to include the invalid tests.
- [x] Add a sufficient number of files for each schema
- [x] Tests should pass


DannyBen:
Tests pass thanks to #468

PR #468, opened 2023-12-20
Add YAML schema as a source to the bashly JSON schema

DannyBen:
An experiment to see if it is worth working with YAML as a basis for generating the `bashly.json` schema.


DannyBen:
Tests are now passing, the schema seems to be valid now.
The only difference is that the custom-properties section is no longer called with `$ref`, and instead is plastered as is multiple times in the JSON file (in the YAML file, it is just once, using YAML anchors).

EmilyGraceSeville7cf (dismissed):

DannyBen:
I squash-merged, and changed the template a little to be more readable.

DannyBen:
@EmilyGraceSeville7cf  - note that I [moved](https://github.com/DannyBen/bashly/commit/a05f7d05edb7041d22b1d8d4db304b36b1eb31df) the YAML source of the schema to the [support folder](https://github.com/DannyBen/bashly/tree/master/support/schema).

PR #476, opened 2023-12-22
Refactor `environment_variable` in schema

DannyBen:


DannyBen:
@EmilyGraceSeville7cf - FYI, I did some changes (in this PR and in https://github.com/DannyBen/bashly/commit/2c75ef0758bde29464a741691cf41e8cd2fc27fb). 

Basically I:

- Removed the "if else" which used to only allow `default` if `required` is false. I tried to re-introduce it without code duplication, but failed, so I moved on.
- Made sure that the environment variable definition appears only once
- Eliminated all the `environment-variables-*` properties from the definitions section, and moved them to their natural place (since there is only one occurrence now).

Another reason I removed the if-else (other than the fact that it was hurting my eyes and adding duplicated code), is that there are other places where some values are allowed in some cases, so it felt inconsistent.

This message is FYI, and in case you have some thoughts on the matter.

Issue #527, opened 2024-07-04
Showing an additional usage line for forced commands


DannyBen:
### Discussed in https://github.com/DannyBen/bashly/discussions/526

<div type='discussions-op-text'>

<sup>Originally posted by **codevogel** July  4, 2024</sup>
Say I have a bashly script `foo` that I run with `./foo`

In my `bashly.yml` I have

```
commands:
  - name: do
    help: do something with my_arg
    default: force
    args:
      - name: my_arg
        help: my argument to be handled
        required: true
```

if I then run `./foo` I get 

```
❯ ./foo
missing required argument: MY_ARG
usage: foo do MY_ARG
```

This could be somewhat confusing to a new user, as they have just done `foo`, not `foo do`.
Also, `./foo MY_ARG` would also be sufficient usage.

So, it would be nice if it instead printed

```
❯ ./foo
missing required argument: MY_ARG
usage: foo do MY_ARG
       foo MY_ARG
```

I don't see how I could achieve this with one of the `custom_strings`, as this would be behavior that is specific to the `forced` command only. 

I tried looking at adding a `src/after.sh` to evaluate the last line of the output and seeing if it starts with `usage: foo do` to echo `usage: foo MY_ARG`, but `after.sh` does not run if the script already exists due to catching a usage error. It also seems like bad practice to do this with a `before.sh`

What would be the best way to enable this?

(An alternative way to render this would be something like)
```
❯ ./foo
missing required argument: MY_ARG
usage: foo MY_ARG
       foo do MY_ARG
```

or 

```
❯ ./foo
missing required argument: MY_ARG
usage: foo do MY_ARG
 ( or: foo MY_ARG )
```</div>

codevogel:
Thanks for the awesome work, I guess this can be closed now?

DannyBen:
Keep open, I will close upon release.

DannyBen:
Version 1.2.0 is released with this implemented.
Thanks for the issue.

PR #530, opened 2024-07-05
Add `show_examples_on_error` setting

DannyBen:
cc #526, #527

This PR adds the ability to specify `show_examples_on_error: true`. When set, the usage message will include the configured examples whenever the user misses a required argument.

```bash
$ ./cli
missing required argument: SOURCE
usage: cli [download] SOURCE TARGET [OPTIONS]
examples:
  cli download example.com
  cli download example.com ./output -f
```


codevogel:
I did find a small bug in this:

If I list no `examples` section, `bashly generate` seems to fail.

with `bashly.yml`
```
name: pls
help: "Project Level Shortcuts - A simple tool to manage project specific aliases."
version: 0.1.0

environment_variables:
- name: pls_default
  help: Path to the root configuration file (The fallback .pls file if is is not found in the current working directory or any of its parents)
  default: $HOME/.pls

commands:
  - name: do
    help: Executes the command associated with the given alias
    default: force
    args:
      - name: alias
        help: The alias to execute
        required: true
```

I get

```
❯ bashly generate
creating user files in src
skipped src/do_command.sh (exists)
 NoMethodError
undefined method `any?' for nil
```

Otherwise the functionality seems great: The examples are listed when I enable the setting. The setting is also added to the defaults generated by `bashly add settings`. The custom string for `examples_caption_on_error` adjusts the output as needed. If I don't have a custom strings file, it just outputs as `examples:` which is the correct casing.

DannyBen:
Excellent report, thanks.
I will fix it and add a test case.

DannyBen:
Fixed.

PR #564, opened 2024-10-22
Update JSON schema to include `command.variables`

DannyBen:
Update schema for #563

### DIFF
--- a/support/schema/bashly.yml
+++ b/support/schema/bashly.yml
@@ -742,6 +742,8 @@ definitions:
         $ref: '#/definitions/sub-command-default-property'
       environment_variables:
         $ref: '#/definitions/environment-variables-property'
+      variables:
+        $ref: '#/definitions/variables-property'
       examples:
         $ref: '#/definitions/examples-property'
       footer:
@@ -788,6 +790,8 @@ properties:
     $ref: '#/definitions/root-version-property'
   environment_variables:
     $ref: '#/definitions/environment-variables-property'
+  variables:
+    $ref: '#/definitions/variables-property'
   examples:
     $ref: '#/definitions/examples-property'
   footer:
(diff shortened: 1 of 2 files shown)


DannyBen:
@EmilyGraceSeville7cf, I have added a new possible value for `command` called `variables`.

It is an array of dictionaries, and each item must have a `name` and an optional `value`. The `value` can be pretty much anything:

```yaml
    variables:
    # Simple value
    - name: output_folder
      value: output

    # Array
    - name: download_sources
      value:
      - youtube
      - instagram

    # Associative array
    - name: zip_options
      value:
        pattern: "*.json"
        compression_level: fast
```

So I did not limit the value to any particular type.
Note that bashly.json schema is auto-generated from the bashly.yaml schema.

Your approval por favor.

support/schema/bashly.yml
```diff
@@ -435,6 +460,15 @@ definitions:
     uniqueItems: true
     items:
       $ref: '#/definitions/environment-variable'
+  variables-property:
+    title: variables
+    description: |-
+      Bash variables for the current application
+      https://bashly.dannyb.co/configuration/command/#variables
+    type: array
+    uniqueItems: true
```
EmilyGraceSeville7cf, line 469: I would also add `minItems: 1`.

support/schema/bashly.yml
```diff
@@ -335,6 +335,31 @@ definitions:
...
+        title: name
+        description: A name for the current variable
+        type: string
+        minLength: 1
+        examples:
+          - output_dir
+      value:
+        title: value
+        description: A value for the current variable
+        examples:
```
EmilyGraceSeville7cf, line 357: What if the variable value is `null` ? Is it an error? What it corresponds to in Bash (in the context of Bashly) if it's not?

EmilyGraceSeville7cf (approved):
LGTM, just have some questions and comments 😄

DannyBen:
Thanks @EmilyGraceSeville7cf - merging.

Issue #592, opened 2024-12-27
Additional information in the default script template header


DannyBen:
### Discussed in https://github.com/DannyBen/bashly/discussions/591

<div type='discussions-op-text'>

<sup>Originally posted by **meleu** December 27, 2024</sup>
## request

I suggest to put in [the default script template header](https://github.com/DannyBen/bashly/blob/master/lib/bashly/views/command/default_script.gtx) an information saying that the code in that file is going to be put **inside a function**.

Something like the second line below:
```bash
echo "# this file is located in 'src/upload_command.sh'"
echo "# the code here will be placed **inside a function**"
```

I'm not really suggesting this exact text, but the message I'd like to be conveyed.

## reasoning

I'm used to put all bash code inside functions. Even for simple scripts, I create a `main` function and then call it with `main "$@"`.

The very first time I was trying bashly, I've coded my `src/root_command.sh` as I would do with my usual way of coding, and it generated this:

```bash
cli_upload_command() {
  # src/upload_command.sh
  hello() {
    echo "hello world"
  }
  hello
}
```

That is actually valid and really works, but the "function inside a function" situation wasn't what I was expecting.</div>

DannyBen:
@meleu - see the pull request - this changes 140 files, so I hope the change is appropriate.

meleu:
Tested here. Working as expected. :+1:

DannyBen:
Thanks, releasing.

PR #593, opened 2024-12-27
Update default script header

DannyBen:
cc #592, #591


meleu:
Tested here and it works as expected! :+1:

Issue #596, opened 2024-12-30
Add support for sourcing the generated script by not running `initialize` and `run`


DannyBen:
*Discussed in https://github.com/DannyBen/bashly/discussions/594*

### Proposed Solution

Change this:
https://github.com/DannyBen/bashly/blob/2073e837f028f9139e63d44ee485a015a66003ed/lib/bashly/views/command/master_script.gtx#L15-L18

to this:

```bash
if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
  initialize
  run "$@"
fi
```

and add this:

```yaml
enable_sourcing: development  
```

to this:

https://github.com/DannyBen/bashly/blob/2073e837f028f9139e63d44ee485a015a66003ed/lib/bashly/libraries/settings/settings.yml#L101-L112

DannyBen:
Version 1.2.8 is released ([changelog](https://github.com/DannyBen/bashly/blob/master/CHANGELOG.md#v128---2025-01-16)) with this change. Thank you all for the feedback and participation.

PR #597, opened 2024-12-30
Add support for avoiding execution if the script is sourced

DannyBen:
cc #596, #594


DannyBen:
@meleu - is this aligned with our discussion?

meleu:
Yep!

Tested here and it indeed generates the initialize/run calls inside the if block by default, and in the top-level scope when `--env production`.

I think now I'm ready to start writing a bashly tutorial! :nerd_face: 

By the way, I hope you have a wonderful 2025! :fireworks:

DannyBen:
> By the way, I hope you have a wonderful 2025! 🎆

You too :)

PR #611, opened 2025-01-20
Update colors library to support auto-colors based on TTY

DannyBen:
cc #600

This PR adds support for auto-color-output based on TTY.
This feature is optional and requires the user to call `enable_auto_colors` in the `initialize.sh`, since it somewhat breaks the binary, stict `NO_COLOR` standard interpretation.

*If this is an acceptable solution, I will fix the specs and update the example/tests accordingly*.


lib/bashly/libraries/colors/colors.sh
```diff
@@ -10,6 +10,14 @@
 ## Color output will be disabled if `NO_COLOR` environment variable is set
 ## in compliance with https://no-color.org/
 ##
+## In case you wish to enable auto detection for color output based on the
+## terminal being interactive (TTY), call `enable_auto_colors` in your 
+## `src/initialize.sh` (Run `bashly add hooks` to add this file).
+##
+enable_auto_colors() {
+  [[ -z ${NO_COLOR+x} && ! -t 1 ]] && NO_COLOR=1 || true
```
pcrockett, line 18: i'll be honest: i love that it's a one-liner, but i really had to write it out and run a few commands to be certain i understood the logic. it's a brain twister.

i kind of hate the following, but i have a feeling my future self would appreciate it more.

```bash
if [[ -z ${NO_COLOR+x} ]]; then
    # NO_COLOR has not been set. Decide whether to output color or not.
    if [[ ! -t 1 ]]; then
        # stdout is not a TTY. Disable color to avoid cluttering output with ANSI escape sequences.
        NO_COLOR=1
    fi
fi
```

pcrockett (approved):
i do think this is a pretty good solution. just one suggestion that i have mixed feelings about.

lib/bashly/libraries/colors/colors.sh
meleu: > i have a feeling my future self would appreciate it more.

Yeah. Exact same feeling here.

Replacing the oneliner with nested if blocks is easier to follow and reason about.

DannyBen: Nested `if` statements with just one leaf are a bit too much in this case.

How about this (with the comment embedded in the lib):

```bash
enable_auto_colors() {
  ## If NO_COLOR has not been set and stdout is not a TTY, disable colors
  if [[ -z ${NO_COLOR+x} && ! -t 1 ]]; then
    NO_COLOR=1
  fi
}
```

meleu: it's good, and removes the `|| true` noise.

meleu (dismissed):

Issue #616, opened 2025-02-03
Cannot properly retrieve INI values with $ signs


le-duane:
### Bashly Version

Latest Ruby Gem

### Description

Hi,

Love Bashly and thank you for sharing it with us. 

I've encountered an issue where my ini values are not getting retrieved correctly when they have `$` embeded in them.

Things are breaking for me on [this line of code:](https://github.com/DannyBen/bashly/blob/966651b3e0c35bd8d14fb9ec8e771d672be82344/lib/bashly/libraries/ini/ini.sh#L50C7-L50C54)

```
[[ $value == *\$* ]] && eval "value=\"$value\""
```

I was hoping you could help me understand its purpose so that I could possibly submit a patch to address and maintain the intended functionality.

Thanks!

### Contents of bashly.yml

```yaml

```

### Reproduction Steps

Embed a dollar sign in any value of your ini e.g.

```
[source]
name = foo$bar
```

pcrockett:
I personally don't use this feature, however it looks like this allows you to specify variables in your INI file, and those variables will get evaluated.

One workaround you could try: stick a backslash (`\`) in front of your dollar signs in your INI file. This way when it gets evaled, it'll come out as a literal dollar sign.

---

For what it's worth, the "bashly version" part of the issue template serves at least two purposes:

* It shows that you've double-checked that you REALLY know what version you're using and aren't just assuming it's the latest.
* It helps our future selves when we look back on this issue.

The latest version today may not be the latest version in 2 weeks, so over time the meaning of "Latest Ruby Gem" changes in meaning. Can you please fill in a specific version number?

le-duane:
Thanks for the feedback - after seeing the history of that file, that does indeed look like what's going on. I have a couple of ideas that I'll submit PRs for.

DannyBen:
Phil is right. This is behaving as expected. Dollar signs denote a variable, you have two options:

1. As mentioned, escape the dollar sign in your INI values.
2. If you don't need this evaluation, remove this line of code or update your copy of the INI library as you see fit. It is created in userland for this purpose exactly.

Thanks for the pull requests, but they both introduce an undesired behavior and complicate the library code.

le-duane:
Sure, they were attempts at addressing the issue while maintaining backwards compatibility. I will proceed as you have suggested.

I would argue however that the best course of action would be to remove the feature entirely or put it behind some kind of flag; since it is non-standard behavior, and possibly more harmful than helpful; as in my case.

[...]

PR #617, opened 2025-02-04
Targeted eval of ini values

le-duane:
For INI config values, check to see if embedded variables exist before substituting their value


DannyBen:
As discussed in #616, closing.

PR #618, opened 2025-02-04
Only substitute INI value for env variable if that variable is set

le-duane:


DannyBen:
As discussed in #616, closing.

Issue #619, opened 2025-02-08
Ability to customize help display behaviour of script name & description via `bashly-strings.yml`


adamscybot:
### Description

When executing the `--help` flag directly on the root script, bashly prints a `<root-name>[ - <root-help>]` string that identifies the script's global name/purpose on the first line of the output. Generally, that makes sense, but I'd like to suppress or customise it in my use case. 

I have a hip, colourful ASCII art preamble that I am currently outputting on _all_ script executions via `initialize.sh`, which I don't care to report how long I spent crafting whilst procrastinating over my core task 😄.

This presents a couple of minor problems:

1. There is a level of duplication when executing `<script> --help` in that the `<root-name>[ - <root-help>]`  preamble string is extraneous in my case since that info (and more) is in my custom header.
2. Whilst I am impressed by my own glorious terminal artwork, I have a feeling I will, in the end, only want it to be displayed on `<script> --help`  and not on every execution in order to recognise the reality of the unappreciative user's sanity :P.

Both problems could probably be solved if this string was customisable in `bashly-strings.yml`. If I only cared about (1) and was happy to keep my preamble on all commands, I could then just null out the hypothetical string template for this in that configuration. If I wanted (2), then I could interpolate some bash into that same string template instead to have my own custom preamble defined there such that it only shows on `<script> --help` .

Worth noting for anyone googling that it doesn't make sense to change/null this out in the `bashly.yaml` root `name` field since that is used for various other operational things and so it's scope is beyond just "display".

An aside: I used this lib some years ago for a whole trough of internal tooling, and it's awesome to see some little challenges I encountered back then have been addressed as I revisit sometime later in an open-source context. Your dedication to this (incredibly time-saving) project that I've noticed whilst lurking on various GitHub threads is admirable! May I suggest a "built with bashly" community-contributed discussion thread/docs page or something?

DannyBen:
Well, isn't this a good looking issue... thanks for taking the time to write it so vividly :).

Let's triage:

### Customizable usage header

Having control over the usage header makes total sense.
However, it has several display modes which makes the update less straight forward than other strings:
https://github.com/DannyBen/bashly/blob/966651b3e0c35bd8d14fb9ec8e771d672be82344/lib/bashly/views/command/usage.gtx#L4-L13
which gets the base string from:
https://github.com/DannyBen/bashly/blob/966651b3e0c35bd8d14fb9ec8e771d672be82344/lib/bashly/script/command.rb#L51-L53

[...]

PR #622, opened 2025-02-15
Update JSON schema with `help_header_override`

DannyBen:
cc #619


DannyBen:
@EmilyGraceSeville7cf can you take a look?

PR #626, opened 2025-03-16
fix quoted here heredocs

rob4226:
Hello,

First off, thank you for creating and maintaining Bashly.  I found it recently and it is an incredibly well-designed tool that has made writing and managing Bash scripts so much easier and more structured. I really appreciate the effort and dedication you’ve put into this project. Keep up the fantastic work!

This PR adds handling for quoted heredocs. Right now, Bashly doesn't recognize quoted heredocs so the generated script fails with syntax errors when executed. 

When the heredoc delimiter is single-quoted or double-quoted, lines in the here-document are not expanded (parameter expansion, command substitution, etc... are ignored). Here is a [link](https://www.gnu.org/software/bash/manual/bash.html#Here-Documents) to the related section in the Bash manual. Also, this [article](https://phoenixnap.com/kb/bash-heredoc#Bash_HereDoc_Syntax) may help explain a bit further.

Link related: #496

### DIFF
--- a/lib/bashly/concerns/indentation_helper.rb
+++ b/lib/bashly/concerns/indentation_helper.rb
@@ -35,7 +35,7 @@ module Bashly
     end
 
     def extract_heredoc_marker(line)
-      line =~ /<<-?\s*(\w+)/ ? $1 : nil
+      line =~ /<<-?\s*['"]?(\w+)['"]?/ ? $1 : nil
     end
 
     def heredoc_closed?(line)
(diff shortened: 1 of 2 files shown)


DannyBen:
Thank you for your kind words, and for taking the time to write this PR and its specs.
My first impression is that it looks good, but we will take a deeper look at it hopefully today.

DannyBen:
Alright, this is simple and good, thank you.

What I don't like - but it is my fault - is that `IndentationHelper` specs are embedded within the `array` specs.
I will merge and move them.

PR #628, opened 2025-04-11
Drop support for Ruby 3.1

DannyBen:


meleu:
Out of curiosity: why?

Some of those dependencies also dropped?

DannyBen:
Because [Ruby 3.1 is EOL](https://endoflife.date/ruby).

The main reasons I am following this practice are:
1. To reduce the time it takes for CI (less versions to check).
2. To open up the possibility of using language features that were added to newer versions of Ruby.

PR #629, opened 2025-04-14
Prevent running validate_* functions twice

meleu:
Currently a [validation function](https://github.com/DannyBen/bashly/tree/master/examples/validations) runs twice:

1. for actually validating
2. only to generate the error message

If validation happens to perform a costly computation, this can be slow and a waste of resources.

This MR makes the validation functions run only once, and then use its output for both validation and error message.


DannyBen (dismissed):
Looks good. I will check it locally tomorrow and merge.
Thanks.

DannyBen:
This is released now in [v1.2.12](https://github.com/DannyBen/bashly/releases/tag/v1.2.12).

PR #630, opened 2025-04-18
avoid conflict with run() function with another bash frameworks

vladriabtsev:
When I am trying to use 'bashmatic' framework for my new bash script, run() function from this framework is conflicting with run() function from bashly generated script

### DIFF
--- a/lib/bashly/libraries/settings/settings.yml
+++ b/lib/bashly/libraries/settings/settings.yml
@@ -126,3 +126,8 @@ var_aliases:
   other_args: ~
   deps: ~
   env_var_names: ~
+
+# Choose different names for some of the internal functions.
+function_names:
+  run: ~
+  initialize: ~
--- a/lib/bashly/settings.rb
+++ b/lib/bashly/settings.rb
@@ -15,6 +15,7 @@ module Bashly
         :enable_inspect_args,
         :enable_sourcing,
         :enable_view_markers,
+        :function_names,
         :lib_dir,
         :partials_extension,
         :private_reveal_key,
@@ -87,6 +88,14 @@ module Bashly
 
       def full_lib_dir
         "#{source_dir}/#{lib_dir}"
+      end
+
+      def function_name(key)
+        function_names[key.to_s] || key.to_s
+      end
+
+      def function_names
+        @function_names ||= get :function_names
       end
 
       def lib_dir
--- a/lib/bashly/views/command/initialize.gtx
+++ b/lib/bashly/views/command/initialize.gtx
@@ -1,6 +1,6 @@
 = view_marker
 
-> initialize() {
+> {{ Settings.function_name :initialize }}() {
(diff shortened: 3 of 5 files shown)


DannyBen:
Thanks for the contribution. I do have a few concerns, mainly around the potential for breaking changes - this could have a significant impact. Also, if I’m remembering correctly, `run` is intended to be available for users to make internal calls as well.

I'm curious about the reasoning behind mixing frameworks here — could you clarify what Bashmatic brings to the table in this context? If there's a use case that truly requires this approach and can’t be addressed in a less disruptive way, I’d definitely be open to exploring it further.

DannyBen:
I pushed a different solution - simply allow customizing the names of some internal functions. Currently, only `initialize` and `run` but this is easily extensible.

It has the advantages of a) being explicit, b) maintaining backwards compatibility.

Thoughts?

vladriabtsev:
I am beginner in writing bash scripts. I have created draft installer script with Bashmatic as framework for UI. I decided to use Bashly to create better parameters for my script. 

Majority Bashmatic functions are grouped and have unique names inside a group. Typical Bashmatic function name is 'group-name.function-name()'. It is helping avoid conflicts when such framework are sourced. At least 'run()' function is exception. It is without group.

Do you know better solution how use several frameworks together and avoid conflicts with function names?

DannyBen:
Thanks for the additional information.

While bashly is designed to be the primary foundation, you should be able to add scripts from other frameworks to it. Here are my thoughts, in general (as I am not fully familiar with bashmatic).

1. Generally speaking, I would try to avoid mixing frameworks - unless, the frameworks are providing solutions for different problems. In the bashly-bashmatic case, I am not sure if this case fits (I would love to be enlightened one way or the other).
2. To my understanding, the main problem you encounter is that the function `run` is used by bashmatic code snippets and other functions to run commands inside a wrapper function. This makes sense, and even without bashmatic in the picture, some people would like to reserve the function name `run` for other uses.
3. I accept the fact that some internal functions - namely `run` and `initialize` - may conflict with other needs.
4. The updated code I pushed to the pull request, addresses this change in a different way - instead of being implicit and not backwards compatible, it is explicit: If someone wants different names, they can easily do so by using bashly settings (`$ bashly add settings` to get the full file).

If this makes sense, I can merge it.

DannyBen:
The schema was not yet updated, but it shouldn't matter.

Your YAML is inorrect, you need indentation and bash function names should not contain a hyphen.

```yaml
function_names:
  run: bashly_run
  initialize: bashly_initialize
```

DannyBen:
I am going to merge this, as I want to work on the schema, and perhaps other related changes. I believe this should solve your use case easily - if you need help, feel free to open a discussion or issue. Thanks for bringing this to our attention.

DannyBen:
This is released now in [v1.2.12](https://github.com/DannyBen/bashly/releases/tag/v1.2.12).

PR #632, opened 2025-06-08
tebako

elee1766:
hi @DannyBen 

im finally following up on this:
https://github.com/DannyBen/bashly/discussions/541

i thought about this again and tried it out once more, and it seems to be somewhat working?

i dont really mean for this to be merged in, its not that well organized or written, but i put these two dockerfiles that seem to compile functional bashly binaries. i also copied one of these binaries to my computer and it seems to run as well. haven't tested it that much more but i thought i would put this here so you/others could see it and maybe try it out

(i also pushed the glibc x86 build by accident but i guess ill.. leave it there)

is there a way to use this compiled binary and test it with the existing rspec tests?


DannyBen:
Nice. My thoughts:

1. The binary itself is - as expected - big, but still smaller than I expected.
2. Testing this executable using existing tests is partially possible - there are "example specs" that test how bashly executable behaves (contrary to the Ruby unit tests) - but still, it would require some tweaking.
3. I for one am not interested in maintaining a packaging solution, unless it is dead simple and cross platform. The maintainer of such a binary packaging solution would have to be able to build it for different platforms, and test it, and support it.
4. If this tebako route would mature, I would consider having it as a separate repository that uses GitHub Actions to build it.

For completeness - could you specify what are the steps/commands needed to build the binary?

DannyBen:
I am closing this. It was a nice idea, but I don't see a future for it.

PR #635, opened 2025-06-21
Add `stacktrace` standard library

DannyBen:
As discussed in https://github.com/DannyBen/bashly/discussions/601

### DIFF
--- a/lib/bashly/libraries/stacktrace/stacktrace.sh
+++ b/lib/bashly/libraries/stacktrace/stacktrace.sh
@@ -20,7 +20,7 @@ stacktrace() {
   printf "%s:%s in \`%s\`: %s\n" "${BASH_SOURCE[0]}" "${BASH_LINENO[0]}" "${FUNCNAME[1]}" "$BASH_COMMAND"
   printf "\nStack trace:\n"
   while caller_output="$(caller $i)"; do
-    read -r line func file <<< "$caller_output"
+    read -r line func file <<<"$caller_output"
     printf "\tfrom %s:%s in \`%s\`\n" "$file" "$line" "$func"
     i=$((i + 1))
   done
--- a/lib/bashly/libraries/stacktrace/stacktrace.sh
+++ b/lib/bashly/libraries/stacktrace/stacktrace.sh
@@ -1,0 +1,28 @@
+## Stack trace functions [@bashly-upgrade stacktrace]
+## This file is a part of Bashly standard library
+##
+## Usage:
+## This function is designed to be called on error.
+## 
+## To enable this functionality, add these lines to your `src/initialize.sh`
+## file (Run `bashly add hooks` to add this file):
+##
+##   trap 'stacktrace' ERR
+##   set -o errtrace
+##
+## Note that this functionality also requires `set -e`, which is enabled by
+## default in bashly generated scripts.
+##
+stacktrace() {
+  local i=0
+  local caller_output line func file
+
+  echo "${BASH_SOURCE[0]}:${BASH_LINENO[0]} in \`${FUNCNAME[1]}\`: $BASH_COMMAND"
+  echo "Stack trace:"
+  while caller_output="$(caller $i)"; do
+    read -r line func file <<< "$caller_output"
+    echo -e "\tfrom ${file}:${line} in \`$func\`"
+    i=$((i + 1))
+  done
+  exit 1
+} >&2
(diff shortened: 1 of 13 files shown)


lib/bashly/libraries/stacktrace/stacktrace.sh
```diff
@@ -0,0 +1,28 @@
...
+  local caller_output line func file
+
+  printf "%s:%s in \`%s\`: %s\n" "${BASH_SOURCE[0]}" "${BASH_LINENO[0]}" "${FUNCNAME[1]}" "$BASH_COMMAND"
+  printf "\nStack trace:\n"
+  while caller_output="$(caller $i)"; do
+    read -r line func file <<<"$caller_output"
+    printf "\tfrom %s:%s in \`%s\`\n" "$file" "$line" "$func"
+    i=$((i + 1))
+  done
+  exit 1
```
meleu, line 27: In my version I made a small improvement about this arbitrary hardcoded `exit 1`.

```bash
stacktrace() {
  local exit_status="$?"

  # stacktrace stuff...

  exit "$exit_status"
}
```

This way we propagate the exit status of the command that caused the crash.

DannyBen: Done.

meleu (approved):

DannyBen:
@pcrockett do you wish to review this before I merge?

pcrockett (approved):
Looks neat. I haven't read the full discussion, but I assume this is pretty thoroughly tested by @meleu over the course of time. Maybe I'll upgrade my tools to use it as well after the next release, see if I can find ways to break it. :+1:

DannyBen:
> Oh, might consider mentioning the [strict setting](https://bashly.dev/usage/settings/#strict) as an alternative to editing `initialize.sh`.

I thought about it, but decided to separate - since you still have to add the `trap` command anyways.

DannyBen:
I could have guessed that trapping error might cause issues in some cases.
I think this is up to the developer to do a conditional trap, something like "if STACKTRACE variable is set" or similar.

meleu:
I think I'll stick with this approach for now and see if it works fine:

Trap `ERR` only if the script is NOT being sourced.

```bash
# initialize.sh

if [[ "${BASH_SOURCE[0]}" != "${0}" ]]; then
  trap 'stacktrace' ERR
fi
```

DannyBen:
> I think I'll stick with this approach for now and see if it works fine

Yup. Anything that works for your use case and audience is valid.

meleu:
@DannyBen one thing that just came to my mind:

The official `stacktrace.sh` could have this:

```shell
enable_stacktrace() {
  trap 'stacktrace' ERR
  set -o errtrace
}
```

That would be similar to what we did with [enable_auto_colors](https://github.com/DannyBen/bashly/blob/21a3f4430c1f488c6d4cfd55343a5cd2e35bc96d/lib/bashly/libraries/colors/colors.sh#L17-L22).

I this allows the `src/initialize.sh` to look a bit tidier. A typical one for my bashly projects would look like this:

```shell
# src/initialize.sh
enable_auto_colors

is_sourced || enable_stacktrace

# whatever...
```

What do you think?

DannyBen:
> @DannyBen one thing that just came to my mind:

Lovely. One thing that bothers me, is the `bashly add hooks` since it adds all hooks, and two of them are "noisy" by default, and have to be deleted by the user. I was toying with the idea of allowing `bashly add hooks initialize` or `bashly add initialize`. to mitigate this and still remain backwards compatible.

Thoughts?

meleu:
Still about refining the stacktrace...

It's quite usual for bash coders to run `bash -x script` for debugging. Even with a stacktrace it's still useful to debug with `-x`.

The thing is that the xtrace's output for `stacktrace` is quite noisy and useless for debugging.

This is how I solved it:

```bash
stacktrace() {
  local exit_status="$?"
  set +x # 👈 MUST be after exit_status assignment
  # ...
}
```

It's important to disable `xtrace` **after** saving the previous exit status, otherwise it would get the exit status of the `set` command.

Do you think it makes sense to make this "official"?

One thing I considered: as the `stacktrace` function involves already has a hard `exit` at the end, I think that "arbitrarily disabling a user-defined bash option" wouldn't be surprising/confusing to the user. On the other hand, the noisy output is quite confusing.

DannyBen:
> It's quite usual for bash coders to run `bash -x script` for debugging. Even with a stacktrace it's still useful to debug with `-x`.

This seems like an over-intervention to me.
I for one, never do that, and I would imagine that if I need to use `-x`, I would want to disable stacktrace.

I feel this falls into the "personal habit" category.

Don't forget that all libraries are dumped into userland for this reason exactly - they serve as a good basis for you to edit if you need to. I do not feel that we need to anticipate the use of `-x`, and even if we do, it is not up to us to override and disable it.

Issue #644, opened 2025-07-28
Multiple consecutive newlines in heredoc are converted to one


m-iwanicki:
### Bashly Version

Latest Docker Image

### Description

Creating heredoc with multiple consecutive newlines (e.g. simulating multiple Enter/default choices) results in those newlines being trimmed to only one newline. This results in possibly hard to debug bugs.

### Contents of bashly.yml

```yaml

```

### Reproduction Steps

```sh
bashly init --minimal
cat <<EOT >src/root_command.sh
fdisk "/dev/sdc" <<EOF
n
p
3


w
EOF
EOT
bashly generate
```

Compare contents of generated download script with `src/root_command.sh`

### Actual Behavior

Only one newline instead of 2.

```sh
grep 'fdisk "/dev/sdc" <<EOF' download -B2 -A7
```

Output:

```sh
root_command() {
  # src/root_command.sh
  fdisk "/dev/sdc" <<EOF
n
p
3

w
EOF

```

Which results in hard to debug failure

### Expected Behavior

The same amount of newlines as in script

```sh
grep 'fdisk "/dev/sdc" <<EOF' download -B2 -A7
```

Output:

```sh
root_command() {
  # src/root_command.sh
  fdisk "/dev/sdc" <<EOF
n
p
3


w
EOF
```

m-iwanicki:
Maybe add visible section in documentation about possible differences in final script that you need to keep in mind e.g. newline trimming (it's also happening in multi-line string). I was quite surprised as I didn't see any mention of this behaviour.
My use case is actually easy to fix e.g.: `printf "%s\n" n p 3 "" "" w | sudo fdisk "/dev/sdc"` which might be more readable than heredoc, not sure about longer examples.

DannyBen:
This can now be mitigated in v1.3.1 by choosing to either disable the formatter, or choosing the external (shfmt) formatter:

```yaml
# Choose a post-processor for the generated script:
# formatter: internal     # Use Bashly's internal formatter (compacts newlines)
# formatter: external     # Run the external command `shfmt --case-indent --indent 2`
# formatter: none         # Disable formatting entirely
# formatter: <string>     # Use a custom shell command to format the script.
#                         # The command will receive the script via stdin and
#                         # must output the result to stdout.
#                         # Example: shfmt --minify
formatter: internal
```

Thanks for the issue.

PR #645, opened 2025-07-29
Fix consecutive newlines in heredoc not being preserved

DannyBen:
cc #644

The referenced issue discovered an edge case with heredocs, where consecutive newlines are being squashed to a single newline.

The solution was to expand the linting operation, from a simple single line that replaced multiple newlines with one, to a full blown `LintHelper` class, similar to the `IndentationHelper` class (for heredoc-aware indentation).

---

## Tasks

- [x] Code implementation
- [x] Tests pass
- [ ] shellcheck and shfmt pass
- [ ] Spec implementation

### DIFF
--- a/lib/bashly.rb
+++ b/lib/bashly.rb
@@ -16,7 +16,7 @@ module Bashly
   ]
 
   autoloads 'bashly/concerns', %i[
-    AssetHelper Completions Renderable ValidationHelpers
+    AssetHelper Completions LintHelper Renderable ValidationHelpers
   ]
 
   module Script
--- a/lib/bashly/concerns/lint_helper.rb
+++ b/lib/bashly/concerns/lint_helper.rb
@@ -1,0 +1,79 @@
+module Bashly
+  # A helper class, used by the `String#lint` extension.
+  # It will remove consecutive newlines and hidden comments from a string
+  # unless the line is within a heredoc block.
+  class LintHelper
+    attr_reader :lines, :output, :marker
+
+    def initialize(script)
+      @lines = script.lines
+      @output = []
+      @marker = nil
+      @previous_blank = false
+    end
+
+    def lint
+      lines.each { |line| process_line line }
+      output.join
+    end
+
+  private
+
+    def process_line(line)
+      if inside_heredoc?
+        handle_heredoc_line line
+      else
+        handle_regular_line line
+      end
+    end
+
+    def handle_heredoc_line(line)
+      output << line
+      reset_marker if heredoc_closed? line
+    end
+
+    def handle_regular_line(line)
+      set_heredoc_state line
+      return handle_blank_line if blank? line
+      return if comment? line
+
+      output << line
+      @previous_blank = false
+    end
+
+    def handle_blank_line
+      return if @previous_blank
+
+      output << "\n"
+      @previous_blank = true
+    end
+
+    def blank?(line)
+      line.strip.empty?
+    end
+
+    def comment?(line)
+      line =~ /^\s*##/
+    end
+
+    def set_heredoc_state(line)
+      @marker = extract_heredoc_marker line unless inside_heredoc?
+    end
+
+    def extract_heredoc_marker(line)
+      line =~ /<<-?\s*['"]?(\w+)['"]?/ ? $1 : nil
+    end
+
+    def inside_heredoc?
+      !!marker
+    end
+
+    def heredoc_closed?(line)
+      line.strip == marker
+    end
+
+    def reset_marker
+      @marker = nil
+    end
+  end
+end
--- a/lib/bashly/extensions/string.rb
+++ b/lib/bashly/extensions/string.rb
@@ -45,7 +45,7 @@ class String
   end
 
   def lint
-    gsub(/\s+\n/m, "\n\n").lines.grep_v(/^\s*##/).join
+    Bashly::LintHelper.new(self).lint
   end
 
   def remove_front_matter
--- a/spec/approvals/examples/stacktrace
+++ b/spec/approvals/examples/stacktrace
@@ -51,5 +51,5 @@ Examples:
 
 Stack trace:
 	from ./download:15 in `root_command`
-	from ./download:260 in `run`
-	from ./download:266 in `main`
+	from ./download:259 in `run`
+	from ./download:265 in `main`


DannyBen:
I would love some input and comments from heavy bashly users or collaborators, to help decide if this PR should be merged or not.

- Merging it means overhead for a niche case of preserving consecutive newlines in heredoc
- Not merging it means heredocs cannot support multiple newlines

@meleu @pcrockett

meleu:
Talking strictly with a **bashly user** point of view: I see what is reported in #644 as an actual bug. Therefore something to be solved.

I would be very disappointed if any behavior coded in a file inside `src/**` wasn't precisely replicated in the final script.

When you mention "overhead", I'm not sure if you're talking about code maintenance cost or increased build time. If it's about the build time, I would happily trade a few hundred milliseconds in script generation time for not being surprised with a situation like the reported one.

If the solution in this MR is not ideal/satisfactory. We can hold it and try to find a better one, but I still think #644 is indeed a bug (maybe more critical than the indentation one, as it changes the behavior).

DannyBen:
Thanks @meleu,

> I would be very disappointed if any behavior coded in a file inside src/** wasn't precisely replicated in the final script.

This is already not precisely replicated, as it is both indented and linted (mainly duplicate newlines removal).

Another possible solution instead of trying to fix the newline linter, is to try and remove the need for it.
Right now, the templates sometimes have a side effect that causes unavoidable sequences of newlines which look quite out of place in the generated script if not handled - I wonder if there is a way to remove the side effect instead of fixing it.

> When you mention "overhead"

Overhead mainly in code maintenance and complexity. This solution adds another - separate - class for handling yet another heredoc edge case.

> If the solution in this MR is not ideal/satisfactory

Yes. It is far from ideal, and intuitively feels like a bloat to me.

---

Perhaps a deeper look into how to eliminate the side effect, which is mainly caused by the ERB templates.

DannyBen:
> I'm a bit confused here... You mean the possibility to use [ERB in config](https://bashly.dev/advanced/erb-in-config/#erb-in-config) is the root cause of this complexity?

No. Bashly uses [GTX templates](https://github.com/DannyBen/gtx), which is just a simple wrapper around ERB, to provide a different syntax more suitable for bashly's templates (Ruby code first instead of template output first). When concatenating several ERB (or GTX) templates together, sequences of newlines form in a way that is not so easy to eliminate.

To see this behavior, one can eliminate the linting function and generate a bashly script. You will see many newlines that are mostly impossible to remove by updating the templates. This is the root cause.

DannyBen:
Here is a minimal demonstration of the problem that the linter is fixing.
I am providing it here as two examples - one plain ERB, and one GTX.

## ERB

```ruby
require 'erb'

insert = <<~TEMPLATE
  sub template
TEMPLATE

subtemplate = ERB.new(insert).result(binding)

template = <<~TEMPLATE
  hello
  <%= subtemplate %>
  <%= another_template if false %>
  <%= another_template if false %>
  world
TEMPLATE

puts ERB.new(template).result(binding)
```

### Output

```
"hello\nsub template\n\n\n\nworld\n"
```

## GTX - this pattern appears a lot in bashly templates

```ruby
require 'gtx'

subtemplate = GTX.new("> sub template").parse

template = <<~GTX
> hello
= subtemplate
= subtemplate if false
= subtemplate if false
= subtemplate if false
> world
GTX

gtx = GTX.new template
p gtx.parse binding
```

### Output

```
"hello\nsub template\n\n\n\nworld"
```

DannyBen:
> do these extra lines affect the behavior of the final script

Not at all, similarly to indentation.
But bashly and its author take great pride at generating a very human readable and aesthetically pleasing script, so it is an important feature.

> Can you give me instructions to achieve this (generate a script with no linting)?

Sure. Assuming you have Ruby 3.2 or higher installed:

```shell
# clone the repo
git clone https://github.com/DannyBen/bashly.git
cd bashly

# update the file `lib/bashly/extensions/string.rb
# look for the `lint` function and replace it with this:
def lint
  self
end

# now you can work inside a folder named "dev" (gitignored)
# and work with bashly almost normally by prefixing it with `bundle exec`
mkdir dev
cd dev
bundle exec bashly init
bundle exec bashly generate
```

> The fictitious scenario I'm wondering here is: if [shfmt](https://github.com/mvdan/sh) was a Ruby gem, would it be ok to remove all this extra-lines cleanup logic and just format the generated script at the end?

This is a great question, even if it was NOT a Ruby Gem.
Should linting remove extra spaces? My view is absolutely. It must. Otherwise it is not linted.

Some linters in some languages do that exactly, even more rigorously - they also add newlines when it is appropriate - even Rubocop in Ruby, and some if not all python linters. One example that comes to mind, in Rubocop, after a `return if ....` statement in Ruby, there must be an empty line if the code continues in the function.

```bash
# this is not linted
somefunc() {
  echo "one";



  echo "two";
}
```

Bottom line - linting implies proper (not too many, not too little) empty lines.

pcrockett:
> The fictitious scenario I'm wondering here is: if [shfmt](https://github.com/mvdan/sh) was a Ruby gem, would it be ok to remove all this extra-lines cleanup logic and just format the generated script at the end?

I think I like where this thought is going. As far as I'm concerned, it doesn't even need to be a Ruby gem. You could just make this an optional dependency. Have something in your docs to the effect of, "If you want a pretty / shfmt-compliant script, make sure you have `shfmt` installed, and Bashly will run it for you automatically."

[I have encountered](https://github.com/Feel-ix-343/markdown-oxide/issues/269) other tools which try to massage code for me, but try to do so with regular expressions or other hackery, without using a proper [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree). This way of doing things is doomed to unending edge case bugs. These tools are infuriating. If we aren't prepared to actually parse the script into an AST before manipulating it, I think the `shfmt` is a much more robust, less infuriating way to do it.

DannyBen:
> If you want a pretty / shfmt-compliant script, make sure you have `shfmt` installed, and Bashly will run it for you automatically.

Hmm. This is interesting, I was unsure that shfmt lints newlines.

While I like the idea of using purpose specific tools, it does feel like we let the tail wag the dog here.
Niche cases of newlines in heredoc, that can be avoided, are the only reason that we cannot simply replace any consecutive newlines with one newline.

DannyBen:
Here is a direction I would like to consider:

1. Revert this sub-par PR back, and go back to simple consecutive newline squasher.
2. Add a bashly setting called `lint` or `linter`:
   ```yaml
   linter: internal  # default behavior
   linter: none / false / no   # no linting at all (other than removing the private comments
   linter: shfmt  # run an internally defined shfmt command
   linter: some_other_command %temp_script_path% --options  # maybe even support this
   ```

Thoughts?

DannyBen:
Closing this, as it was fixed in a different way in #646.

PR #646, opened 2025-08-01
Add ability to choose a different formatter

DannyBen:
cc #644, #645

This PR comes to replace #645, and solve the consecutive heredoc newlines using a different approach.

1. There is now a new setting: `formatter: internal`.
2. Setting it to `none` will avoid compacting newlines completely.
3. Setting it to `external` will format it using `shfmt --case-indent --indent 2`.
4. Setting it to any other string (e.g. `shfmt --minify`) will assume this is the command to run, and run it as a formatter. The command will be sent the pre-formatted script through stdin, and is expected to print the formatted script to stdout.

The previous `String#lint` extension method was split into two:
1. The new `String#remove_private_comments` which removes the `## comments` from the script
2. The newline compaction responsibility is now in a new `Script::Formatter` class

---

## TODO

- [x] Proof of concept
- [x] Finalize implementation
- [x] Existing tests pass
- [x] New code has new specs
- [x] 100% spec coverage
- [x] Documentation - [Source](https://github.com/DannyBen/bashly-book/pull/43) or [Preview](https://1-3.bashly-book.pages.dev/usage/settings/#formatter)


DannyBen:
Thanks for the test. I am pushing a version in a few that passes all specs in my Ruby 3.4, and we can then see if it passes in other Rubys. I can't tell anything from the ArgumentError you observed, other than the fact that this is an internal Ruby error, and not a bashly / shfmt error - so perhaps differences between our ruby versions - which one you use?

meleu:
```
$ ruby --version
ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [x86_64-linux]

$ shfmt --version
3.12.0
```

lib/bashly/script/formatter.rb
```diff
@@ -0,0 +1,40 @@
...
+    private
+
+      def shfmt_result
+        unless system 'command -v shfmt > /dev/null 2>&1'
+          raise DependencyError,
+            'Cannot find g`shfmt`.\nEither install it or change to the g`internal` formatter.'
+        end
+
+        output, error, status =
+          Open3.capture3 %w[shfmt --case-indent --indent 2], stdin_data: script
```
meleu, line 32: I suggest using only `shfmt` here.

Reason being that I have my own preferences for `shfmt` defined in an `.editorconfig` file. And according to `man shfmt`:

> If any EditorConfig files are found, they will be used to apply formatting options. **If any parser or printer flags are given to the tool, no EditorConfig formatting options will be used.**

DannyBen: > I suggest using only `shfmt` here.

Hmm. Most people wont have your setup, and I want them to have a script that is formatted in the same way bashly scripts were formatted up until now, which needs these options.

What I can do, is allow custom command, which will allow any command as long as it accepts input as stdin, and prints output to stdout.

Then all "internal" options will be changed to require a colon prefix to distinguish between the symbols and any other string, will be treated as command. 

This should allow you to to do `formatter: shfmt` or `formatter: shfmt --any --option`

Yes?

meleu: I'm not really sure here... :thinking: 

I don't like the `:symbol-like` notation in the yaml file. It can be familiar for us, Ruby people, but looks unconventional in a yaml file. I can imagine users being confused with it (and requesting contributor's time opening issues/discussions to talk about this).

In the face of making people having to use Ruby symbol-like notation, I withdraw my suggestion.

Although I justified my request saying that I have a custom `.editorconfig`, the only situation where I open the final script for reading is when I want to debug something (thanks to our lovely stacktrace :slightly_smiling_face:). And for debugging I'm happy with any `shfmt`ed (whatever options being used).

---

On the other hand, I can also imagine hardcore users opening issues/discussions with a request to make `shfmt` respect their `.editorconfig`. But these kind of users are much less numerous than the regular ones.

DannyBen: Too late. I can see reasons where users want their own `shfmt` options, or even a whole different formatter.

The support for custom command is done. I don't see the symbol values as a big deal - they are fully documented in the default settings.yml, but if there is a better option for using "known strings" and any unknown string is a custom command, I am definitely in favor.

The problem is that just "shfmt" may mean clean shfmt (no options) or the internal shfmt (with my default options).

DannyBen: How about:
- internal
- external (my flavor of shfmt)
- none
- any other string = custom command

Boom?

meleu: I like this last 'non-ruby-symbol' notation proposal. However I would suggest something like `bashly-shfmt` instead of `external`, as it would be more clear.

What do you think?

meleu:
I'll be available for careful testing over this weekend.

lib/bashly/script/formatter.rb
meleu: Let's go with this approach then. The explanation is clear enough :+1:

DannyBen:
@pcrockett since it was you who nudged us in this direction, if you have time to review, I am sure it will be beneficial.

pcrockett:
in any case, the feature as described by [the docs](https://github.com/DannyBen/bashly-book/pull/43) sounds great. :+1:

DannyBen:
> it might take me a few days to get around to thoroughly reviewing

No pressure. I will wait for several days at least as I prefer merging after your input than having to change it later.
Since it is a bigger than usual feature, I also assigned the new version 1.3 instead of 1.2.x.

spec/bashly/script/formatter_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      it 'raises a Bashly::Error' do
+        expect { subject.formatted_script }.to raise_approval 'formatter/error-not-found'
+      end
+    end
+
+    context 'when the external command fails' do
+      let(:mode) { 'shfmt --diff' }
+
+      it 'raises a a Bashly::Error' do
```
meleu, line 48:
```suggestion
      it 'raises a Bashly::Error' do
```

meleu (approved):
I'm happy with the results of this PR.

Nice job! :+1:

Issue #643, opened 2025-07-28
[Feature / Question] Access to exact raw args?


mcblum:
### Description

Hello again!

I'm working on implementing a small CLI to assist my family in configuring their Ubiquiti products so that they "just work". One of the quirks of this process is that sometimes you need to change the user's group when running a command when changing the config on the device. The process they suggest is to basically re-execute the command you just ran but with a different group. Doing a little investigation, it appears to me that Bashly has already processed some things about the command args even when initialize.sh runs - is that correct?

This is the code snippet that they recommend as the way to re-execute the command but with a different group:
```
if [ "$(id -g -n)" != 'vyattacfg' ] ; then
    exec sg vyattacfg -c "/bin/vbash $(readlink -f $0) $@"
fi
```

I can, of course, add some config to each command and re-construct the exact command that was run, but, before I do that, is there a way to access the raw args within the Bashly structure?

Thank you, the new features added recently are dope.
Matt

pcrockett:
in hindsight, as long as `input` is an undocumented unintentional feature, it might be wise to change those kinds of variable names so they are harder to accidentally overwrite. ex: `__BASHLY_INPUT` or something like that. i can totally see myself defining an `input` variable in a before hook for whatever reason, then being surprised when things stop working.

DannyBen:
The decision to avoid namespacing global variables and internal functions (think `inspect_args()`), along with the decision to use as few of them as possible, was very intentional. As a user, I am working on bash scripts, I do not want to be constantly made aware of the framework and its needs. 

If the word "bashly" appears in a generated script (except the optional info header), I consider this a design flaw.

That said - perhaps it would be a good idea to document all the different variables or functions that may be considered "reserved words" in a bashly environment. 

The thing is that over-documenting becomes pointless at a certain level. I mean, if someone happens to bump into this issue, it will be easier for them to look inside the script and identify the problem than to go to some documentation site and find out where the documentation author chose to place it, if at all.

DannyBen:
@pcrockett another thing I just observed: The `input` array that is accessibly in the `before` hook has already served its purpose, so even if you use this name anywhere, it should not be a problem.

This code snippet from `run()` shows why:

```bash
  normalize_input "$@"
  # consume `input` here and convert it to `args`
  parse_requirements "${input[@]}"
  # from this point it is no longer used
  before_hook
```

pcrockett:
> sort of a symbiotic relationship between the developer and bashly.

[...]

PR #654, opened 2025-08-06
Add ability to override raw input

DannyBen:
cc #643

### DIFF
--- a/examples/hooks/src/initialize.sh
+++ b/examples/hooks/src/initialize.sh
@@ -3,4 +3,10 @@
 ## Any code here will be placed inside the `initialize()` function and called
 ## before running anything else.
 ##
+## - The raw command line arguments are available to you here as "$@"
+## - If you wish to override the raw arguments before they are processed,
+##   simply define an array named `input_override`:
+##
+##     declare -g -a input_override=("modified" "args")
+##
 ## You can safely delete this file if you do not need it.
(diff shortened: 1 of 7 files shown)


DannyBen:
@mcblum @pcrockett 
This is the best way I found to allow advanced developers to "hack" the raw input line, and replace it if they want.
It allows replacing it in the `initialize`, by declaring an array named `input_override`.

@meleu
if you wish to chime in, you are welcome as usual

DannyBen:
Will this solve your use case?

pcrockett:
hm, i dunno. i think the VAST majority of scripts out there have no need to manipulate raw args. so adding a logic branch just for this one edge case seems a bit wonky to me.

i tend to lean more toward something like this:

```bash
raw_args="$@"
initialize # has ability to manipulate raw_args if desired
run "${raw_args[@]}"
```

this wouldn't clutter up my scripts so much, but would give "edge-case" folks the flexibility they want. wdyt?

DannyBen:
> so adding a logic branch just for this one edge case seems a bit wonky to me.

Yes, thank you. Totally agree.
So you do not object to the idea, just the implementation - I will update it to match yours.

pcrockett (approved):
Looks good to me. Small nit, not blocking a merge:

I'm not an enormous fan of the `command_line` name, as that implies to me that it includes the full command (including the script name / path). Something with "args" in the name (i.e. `command_line_args`) makes more sense to me. It's a matter of preference / taste though.

And other by-the-way comments:

* I'm not sure how you like to handle PRs, but you might consider updating the original description with the current state of the PR, so future people don't need to read all the way through to know what actually got merged.
* I actually unsubscribed from this PR because I was getting too many notifications (every time you pushed a commit -- not your fault, just GitHub's). I luckily checked it again today and saw you asked me for another look. In the future, (re-)requesting a review from me would be more likely to get my attention.

Issue #659, opened 2025-08-09
[minor] generated root_command.sh mentions wrong function name


meleu:
### Bashly Version

Latest Ruby Gem

### Description

We can see an example [here](https://github.com/meleu/rndm/blob/ae718c1434bbf2e1fe95bde8b570f7c125b87718/rndm#L13-L21):

<img width="793" height="332" alt="Image" src="https://github.com/user-attachments/assets/780fb822-98d2-40f6-a8be-2cb889940dd1" />

(finding these minor things is a \[positive] side-effect of the tutorial writing endeavor I'm currently involved 🙂)

### Contents of bashly.yml

```yaml

```

### Reproduction Steps

```bash
cd examples/minimal/
bashly generate
```

Inspect the generated `download` script and you'll see that the `root_command()` function mentions that the code will be wrapped by a function named **`download_command()`**, which is incorrect.

DannyBen:
Confirmed. Fixing.

DannyBen:
Fixed in 1.3.2.

PR #660, opened 2025-08-10
Fix function name in root command comment

DannyBen:
cc #659

The root command behaves slightly differently than other commands, in terms of naming.

Other commands - say `download` - have all their functions prefixed by the name of the command - `download_command`, `download_usage` etc.

The root command has a split behavior: The command function is named `root_command`, while other functions are prefixed by the CLI name - `download_usage`.

Due to this behavior, we cannot just update the `Script::Command.function_name` method:

https://github.com/bashly-framework/bashly/blob/14f1c944b754e797096f229d8002bad146b08a2a/lib/bashly/script/command.rb#L67-L69

and instead, we have to address this particular case with a local template condition:

```bash
> echo "# The code you write here will be wrapped by a function named '{{ root_command? ? 'root' : function_name }}_command()'."
```

### DIFF
--- a/examples/catch-all/src/root_command.sh
+++ b/examples/catch-all/src/root_command.sh
@@ -1,5 +1,5 @@
 echo "# This file is located at 'src/root_command.sh'."
 echo "# It contains the implementation for the 'download' command."
-echo "# The code you write here will be wrapped by a function named 'download_command()'."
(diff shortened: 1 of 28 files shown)


meleu (approved):
Confirmed it's working as expected.

Issue #658, opened 2025-08-07
Consider supporting array configuration in validations


DannyBen:
As discussed in #651

meleu:
Hey @DannyBen

Real Life got in my way, 3 weeks passed and I still wasn't able to finish this...

I started taking a look at it today but then some questions paralyzed me. 

Should we support both string and array? (just like we do for alias and examples.) The reason is that making it "array only" would be a breaking change, no?

Or maybe provide a brand new feature called `validations` (plural) supporting an array?

Are we going to provide both `validate` and `validations`? Or deprecate `validate` in favor of `validations` (for consistency sake)?

What are your thoughts on this?

Can you pick this (so I can put my focus back in the tutorial)?

---

By the way, can you release 1.3.2 with the #660 naming fix?

I understand this is a minor thing, in a comment that will be deleted by the user. But it's just so I can post the part 1 of the tutorial with an accurate information.

PR #663, opened 2025-08-28
Add array syntax to `validate`

DannyBen:
cc #658


meleu (approved):
In my tests it works exactly as I requested. Thanks! :heart:

PR #669, opened 2025-08-30
Add support for additional settings file names

marlondecol:
This pull request proposes support for additional names for the settings file.

In addition to `bashly-settings.yml` and `settings.yml`, Bashly would also support `bashlyrc.yml`, as well as dotted versions of all three. This means it would recognize the following files, in this order, as the settings file:

- `.bashly-settings.yml`
- `bashly-settings.yml`
- `.bashlyrc.yml`
- `bashlyrc.yml`
- `.settings.yml`
- `settings.yml`

Note that `bashly.yml` (and its dotted version) was intentionally excluded to avoid confusion with the main configuration file.

This naming format tries to follow a convention adopted by other tools. While it is not a strict rule, I believe other developers might appreciate the added flexibility.

These naming choices are arbitrary. Any feedback or suggestions are welcome.

---

_This was my first experience with Ruby, so I hope this doesn't cause any issues!_ 🙂

### DIFF
--- a/lib/bashly/settings.rb
+++ b/lib/bashly/settings.rb
@@ -181,13 +181,12 @@ module Bashly
       end
 
       def user_settings_path
-        @user_settings_path ||= if ENV['BASHLY_SETTINGS_PATH']
-          ENV['BASHLY_SETTINGS_PATH']
-        elsif File.exist? 'bashly-settings.yml'
-          'bashly-settings.yml'
-        else
-          'settings.yml'
-        end
+        @user_settings_path ||= ENV['BASHLY_SETTINGS_PATH'] || [
+          '.bashly.yml', 'bashly.yml',
+          '.bashly-settings.yml', 'bashly-settings.yml',
+          '.bashlyrc.yml', 'bashlyrc.yml',
+          '.settings.yml'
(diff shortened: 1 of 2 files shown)


DannyBen:
Hey Marlon,

Thank you for your contribution. I appreciate the time and effort you put into this PR, especially since you mentioned that this is your first experience with Ruby - you've done a great job with it.

That said, I won't be able to merge this particular PR. The reason is that the project's settings file mechanism is intentionally limited to `settings.yml` and `bashly-settings.yml`. Bashly is designed to work in a structured directory workspace, so straightforward filenames like `settings.yml` make sense. Filenames such as `*rc` files or hidden dotfiles don't fit as well with this approach, and I'd like to keep the scope focused and consistent.

For future changes that might affect core behavior, it's usually best to open an issue or start a discussion first. That way we can talk through the approach together before diving into implementation.

I hope this won't discourage you from contributing again in the future.

marlondecol:
No problem! I understand that sometimes a proposal may not be accepted or a change may not be implemented — that’s part of the process. I also understand and respect your arguments and the scope constraints.

I’ll make sure to open an issue first whenever I have a new contribution to bring. Thanks a lot for your feedback!

PR #718, opened 2026-07-09
Add support for negatable flags (`--[no-]color`)

DannyBen:
The main motivation behind this feature, is the recently added [argfile support](https://bashly.dev/configuration/command/#argfile), which means that users can define something like `--color` in their argfile, and then they have no way to disable it.

This can now be mitigated easily with:

```yaml
flags:
- long: --color
  short: -c
  help: Enable color output
  negatable: true
```

---

- [x] Implementation
- [x] Specs
- [x] Bashly bash completions
- [x] Bashly CLI docs (`bashly doc flag.negatable`)
- [x] JSON schema
- [x] bashly.dev docs - [Source](https://github.com/bashly-framework/bashly-book/commit/ea1b6e2964452e7c3448a55cc54b94ba9ca07e19) | [Preview](https://1-4-0.bashly-book.pages.dev/configuration/flag/#negatable)


meleu:
You'll receive a notification because of this message and I really hope you don't take it as a waste of your time.

Instead, I want you to receive my appreciation of your work on this feature! That's going to decrease the amount of code we need to handle just to add the `--no-` version of some options.

Thanks dude! :heart:

DannyBen:
@meleu, are you kidding buddy? I thought we were friends... :)

a) I am always happy to hear that someone else finds things I do useful, even more so from long time users and contributors.
b) If you wanted this feature, you could have asked

Personally, I didn't see a point to this feature before we added the argfile support - now users can state `--color` in the argfile, and then there is no way to turn it off, so `--[no-]color` now has a stronger reasoning for existing.

I am curious what is your use case for needing it.

